PR #14, opened 2019-02-24
Add color param to Text View

TobseF:
Let's add a `color` param to the `Container.text(...)`. So we can change the color of texts.


soywiz:
Thanks!

Issue #47, opened 2019-12-02
Scaling issues with different build targets


emign:
Building the exact same project for different targets results in different view scalings:

korge-plugin-version: 1.4.2 (JS targt from 1.4.1)

First image: MacosnativeX64
Second image: JVM (still AWT/Swing, not JNA)
Third: JS
<img width="1392" alt="Bildschirmfoto 2019-12-02 um 19 58 50" src="https://user-images.githubusercontent.com/5278352/69987723-6da76a80-1540-11ea-869d-08c4849de7b1.png">
<img width="1392" alt="Bildschirmfoto 2019-12-02 um 19 55 36" src="https://user-images.githubusercontent.com/5278352/69987724-6da76a80-1540-11ea-8cf6-27e21a08dc99.png">
<img width="952" alt="Bildschirmfoto 2019-12-02 um 20 10 32" src="https://user-images.githubusercontent.com/5278352/69987726-6e400100-1540-11ea-8c7f-a9cafda33e54.png">

emign:
Add JVM windows to the list

emign:
Only an update sincd 1.4.c now is live: It happens with the JNA target too
So the current state is:
JS/Browser: Fullscreen (but images get pixelated)
JNA/JVM/Native: Seems half width and half height

<img width="1472" alt="JS" src="https://user-images.githubusercontent.com/5278352/70600032-f41e1500-1bee-11ea-9010-e18c938c7b90.png">
<img width="1348" alt="JNA" src="https://user-images.githubusercontent.com/5278352/70600037-f6806f00-1bee-11ea-8369-c2431710b6fd.png">

soywiz:
Note to myself: probably related to having the main display as retina and displaying the window in a secondary display without retina. So I will have to check the pixel device ratio when moved/resized

soywiz:
I have hopefully fixed all the targets: `runJvm`, `runJvmFirstThread` and `runNative*` on mac. For the record, and as talked at slack. To validate this before the next version:

KorGE uses mavenLocal `~/.m2` folder first instead of jcenter/bintray. So if you clone korgw repo, and changes the `gradle.properties` `version=1.9.7.SNAPSHOT` to `version=1.9.6` and then install, you can try the changes with Korge `1.4.3c` that references that version locally.

To try it faster since native and other targets might take some more time to build. Let's do this first:

Clone the repo, change the gradle.properties, and then execute:
* `./gradlew publishJvmPublicationToMavenLocal`
[...]

PR #49, opened 2019-12-08
Convert Pixel to Worldcoordinates

emign:
If theres a better way to find the worldView of a stage besides
```
 children.getOrNull(0)
```
maybe one could also get rid of the nullcheck

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/Views.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/Views.kt
@@ -311,8 +311,7 @@ fun viewsLog(callback: suspend Stage.(log: ViewsLog) -> Unit) = Korio {
 	callback(log.views.stage, log)
 }
 
-open class GameWindowLog : GameWindow() {
-}
+open class GameWindowLog : GameWindow()
 
 class ViewsLog(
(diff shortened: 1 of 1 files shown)


soywiz:
Sorry for the delay, I missed this PR.

If its referencing box2d shouldn't it be an extension inside box2d project?

Doesn't `View.localToGlobal` or `View.localMouseX` work for you in this case?

emign:
I will check the placement. I dont fully understand what `View.localToGlobal` or `View.localMouseX` are doing at the moment. Box2d uses another coordinate system, where the origin is in the center and it goes negative to the right. The View methods dont seem to use that.
` println(convertPixelToWorld(Point(view.width/2,view.height / 2)))` should give (0,0) as result
ch05_02

soywiz:
Oh I see, then please move the extensions here (since only makes sense when using box2d) and I'll merge it:
https://github.com/korlibs/korge/blob/master/korge-box2d/src/commonMain/kotlin/com.soywiz.korge.box2d/Box2d.kt

emign:
All right. I will do after we worked out [47](https://github.com/korlibs/korge/issues/47) because debugging/testing will be much easier then

PR #54, opened 2019-12-11
Box2D and Korge improvements

soywiz:
Box2D improvements:
- Documentation for WorldView and classes
- New createSimpleBody
- RectBase.anchor, center and centered extensions
- Allow to debug views printing lines around shapes

Korge improvements:
- RenderContext plugable generic flusher
- New DebugLineRenderContext (RenderContext.debugLineRenderContext) allowing to draw lines


soywiz:
Also now setting the view, creates always an internal Container

PR #56, opened 2019-12-12
Added pixel and world conversion functions to WorldView

emign:
Added to Worldview because its a view operation


soywiz:
Thanks!

PR #62, opened 2020-01-03
Customizable UiViews

RezMike:
Make UiViews customizable.


korge/src/commonMain/kotlin/com/soywiz/korge/newui/UIComboBox.kt
```diff
@@ -5,6 +5,7 @@ import com.soywiz.korge.view.*
 import com.soywiz.korim.color.*
 import com.soywiz.korio.async.*
 
+@Deprecated("Use .ui.uiComboBox instead")
```
soywiz: What about adding `@Deprecated("Use .ui.uiComboBox instead", replaceWith = ReplaceWith("..."))`

korge/src/commonMain/kotlin/com/soywiz/korge/newui/UIComboBox.kt
```diff
@@ -14,6 +15,7 @@ inline fun <T> Container.uiComboBox(
 	block: UIComboBox<T>.() -> Unit = {}
 ) = UIComboBox(width.toDouble(), height.toDouble(), selectedIndex, items, skin).also { addChild(it) }.also(block)
 
+@Deprecated("Use .ui.UIComboBox instead")
```
soywiz: Could it be a typealias with Deprecated? If not, since newui is pretty new, maybe we could just delete these clases

korge/src/commonMain/kotlin/com/soywiz/korge/newui/UIComboBox.kt
RezMike: Hm, good suggestion. I'll try to add this.

korge/src/commonMain/kotlin/com/soywiz/korge/newui/UIComboBox.kt
RezMike: I don't know about typealiases since there may be compile time errors after updating the korge dependency, if there are developers who used the old api in their projects.

RezMike: But if there are no such developers, I would delete the old classes too.

korge/src/commonMain/kotlin/com/soywiz/korge/ui/UICheckBox.kt
```diff
@@ -0,0 +1,101 @@
...
+import com.soywiz.korma.geom.*
+
+inline fun Container.uiCheckBox(
+	width: Number = 120.0,
+	height: Number = 32.0,
+	checked: Boolean = false,
+	text: String = "CheckBox",
+	textFont: Html.FontFace = defaultUIFont,
+	skin: UISkin = defaultUISkin,
+	checkedSkin: UISkin? = null,
```
soywiz: I would totally drop the X text and would include a texture slice for the checkbox in its states as part of the UISkin itself

soywiz: Still we can do this later, the parameter

korge/src/commonMain/kotlin/com/soywiz/korge/ui/UIComboBox.kt
```diff
@@ -0,0 +1,142 @@
...
+	selectedIndex: Int = 0,
+	items: List<T>,
+	verticalScroll: Boolean = true,
+	textFont: Html.FontFace = defaultUIFont,
+	skin: UISkin = defaultUISkin,
+	showSkin: UISkin = skin,
+	hideSkin: UISkin? = null,
+	upSkin: UISkin? = null,
+	downSkin: UISkin? = null,
+	scrollbarSkin: UISkin = skin,
```
soywiz: Why not include the colors/images used for the UI as part of a single UISKin?

korge/src/commonMain/kotlin/com/soywiz/korge/ui/UIScrollableArea.kt
```diff
@@ -0,0 +1,108 @@
...
+	contentHeight: Number = 512.0,
+	buttonSize: Number = 32.0,
+	verticalScroll: Boolean = true,
+	horizontalScroll: Boolean = true,
+	font: Html.FontFace = defaultUIFont,
+	skin: UISkin = defaultUISkin,
+	upSkin: UISkin? = null,
+	downSkin: UISkin? = null,
+	leftSkin: UISkin? = null,
+	rightSkin: UISkin? = null,
```
soywiz: Ditto. Why not use a single UISkin and allow there to provide specific skins defaulting to the basic ones? This would allow to simplifying the propagation of the skin. Maybe you can use a typed mechanism like the coroutines context to provide ways to extend the available skin stuff and provide method/property extensions in this file like:
```fun UISkin.withScrollLeftButton(...): UiSkin```

File: korge/src/commonMain/kotlin/com/soywiz/korge/view/Container.kt [hunk 6]
```diff
@@ -26,7 +26,7 @@ inline fun Container.container(callback: @ViewsDslMarker Container.() -> Unit =
  */
 @UseExperimental(KorgeInternal::class)
 open class Container : View() {
-	override val isContainer get() = false
+	override val isContainer get() = true
```
soywiz: Woops :)

korge/src/commonMain/kotlin/com/soywiz/korge/ui/UICheckBox.kt
RezMike: I would drop the X text too, but the DefaultUISkin doesn't have any alternative for the X sign, so the checked state of the default implementation of uiCheckBox (with defaultUISkin) will not be displayed correctly. So in order to drop it, we need to provide a correct default skin.

korge/src/commonMain/kotlin/com/soywiz/korge/ui/UIComboBox.kt
RezMike: We can create different skins for different views, so each view skin will have a unique set of parameters to specify its own state, it's a quite good solution. But what about the case when a developer want each component of a view (for example, each button of uiComboBox) to look different? In this case, I suppose, we need different skins for each component.

korge/src/commonMain/kotlin/com/soywiz/korge/ui/UIScrollableArea.kt
RezMike: What about creating additional data class ScrollableAreaSkin that will include these properties with default values: skin, upSkin, downSkin, leftSkin, rightSkin?

RezMike:
For now, I am thinking about different complex skins for each view. Like I commented above: "data class ScrollableAreaSkin that will include these properties with default values: skin, upSkin, downSkin, leftSkin, rightSkin"

soywiz:
Thanks! I have merged it. Lets cover the remaining stuff separately.
We can extract the bytes from the base64 string, add the checkbox stuff then save again, run ImageOptim and re-encode as base64

RezMike:
What about additional skins for complex views as I mentioned above? I can create a preview PR with that.

PR #72, opened 2020-01-19
Add amplitude to NativeVibration

TobseF:
Without JDoc the vibration usage is hard to understand and behaves different on JS and Android. And it always uses the maxium vibration amplitude.
For example to vibrate 200ms on Android, you had to write:
``` kotlin
vibrate(longArrayOf(0, 42))
```

**Changes**
* Add `vibrate(time: TimeSpan, amplitude: UInt)`
  * Add single vibration
  * Add amplitude for Android target (_amplitude intensity of the vibration [0-255])_
* Renamed `vibrate(..)` to `vibratePattern` - nessesary to resolve name clash on expect/actual matching
* Renamed pattern to timings
* Fix pattern usage:
   Pattern is always a list of timings for ON and OFF sequences. But on JS it starts with ON, on Android 
  with OFF. So this fix adds a empty off timing for Android.
* Add JDoc for vibration methods

It's discussible if `amplitude` should be
1. A Int range from 0 - 255
2. A UInt range from 0 - 255
3. A Int range from 0 - 100
4. A Double range from 0 - 1

I have chosen 2. because it's the range on Android and it prevents negative values.

I had problems to build korge, because of:
> Expected class 'NativeVibration' has no actual declaration in module
> Expected object 'NativeStorage' has no actual declaration in module

So I only tested the vibration on a separate android project.

### DIFF
--- a/korge/src/androidMain/kotlin/com/soywiz/korge/service/vibration/NativeVibration.kt
+++ b/korge/src/androidMain/kotlin/com/soywiz/korge/service/vibration/NativeVibration.kt
@@ -7,6 +7,8 @@ import android.support.v4.content.ContextCompat.getSystemService
 import com.soywiz.klock.TimeSpan
 import com.soywiz.korge.android.androidActivity
 import com.soywiz.korge.view.Views
+import kotlin.math.abs
+import kotlin.math.min
 
 actual class NativeVibration actual constructor(views: Views) {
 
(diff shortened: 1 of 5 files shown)


korge/src/androidMain/kotlin/com/soywiz/korge/service/vibration/NativeVibration.kt
```diff
@@ -1,16 +1,46 @@
...
+    /**
+     * @param time vibration duration in milliseconds
+     * @param amplitude intensity of the vibration (0-255)
+     */
+    @ExperimentalUnsignedTypes
     @SuppressLint("MissingPermission")
-    actual fun vibrate(pattern: Array<TimeSpan>) {
-        val v = getSystemService(views.androidActivity, Vibrator::class.java)
-        v?.vibrate(VibrationEffect.createWaveform(pattern.map { it.millisecondsLong }.toLongArray(), -1))
+    actual fun vibrate(time: TimeSpan, amplitude: UInt) {
```
soywiz: If it is a value between 0-255 why not use a UByte instead, or just an Int?
Any why not use a Double or Float with a value between 0.0 and 1.0?

korge/src/androidMain/kotlin/com/soywiz/korge/service/vibration/NativeVibration.kt
```diff
@@ -1,16 +1,46 @@
...
+
+    private val vibrator = getSystemService(views.androidActivity, Vibrator::class.java)
+
+    /**
+     * @param timings list of alternating ON-OFF durations in milliseconds. Staring with ON.
+     * @param amplitudes list of intensities of the vibration (0-255)
+     */
+    @ExperimentalUnsignedTypes
+    @SuppressLint("MissingPermission")
+    actual fun vibratePattern(timings: Array<TimeSpan>, amplitudes: Array<UInt>) {
```
soywiz: Ditto about using Float/Double for amplitude

TobseF:
Closed because of a changed and squashed:
https://github.com/korlibs/korge/pull/73

PR #73, opened 2020-01-20
Add amplitude to NativeVibration

TobseF:
Without JDoc the vibration usage is hard to understand and behaves different on JS and Android. And it always uses the maxium vibration amplitude.
For example to vibrate 200ms on Android, you had to write:
``` kotlin
vibrate(longArrayOf(0, 42))
```

**Changes**
* Add `vibrate(time: TimeSpan, amplitude: UInt)`
  * Add single vibration
  * Add amplitude for Android target (_amplitude intensity of the vibration [0-255])_
* Renamed `vibrate(..)` to `vibratePattern` - nessesary to resolve name clash on expect/actual matching
* Renamed pattern to timings
* Fix pattern usage:
   Pattern is always a list of timings for ON and OFF sequences. But on JS it starts with ON, on Android 
  with OFF. So this fix adds a empty off timing for Android.
* Add JDoc for vibration methods

It's discussible if `amplitude` should be
1. A Int range from 0 - 255
2. A UInt range from 0 - 255
3. A Int range from 0 - 100
4. A Double range from 0 - 1

I have chosen 4. because of carlos 😉.

I had problems to build korge, because of:
> Expected class 'NativeVibration' has no actual declaration in module
> Expected object 'NativeStorage' has no actual declaration in module

2020-01-20_20h52_08

So I only tested the vibration on a separate android project.

### DIFF
--- a/korge/src/androidMain/kotlin/com/soywiz/korge/service/vibration/NativeVibration.kt
+++ b/korge/src/androidMain/kotlin/com/soywiz/korge/service/vibration/NativeVibration.kt
@@ -1,16 +1,56 @@
 package com.soywiz.korge.service.vibration
 
-import android.annotation.*
-import android.os.*
-import android.support.v4.content.ContextCompat.*
-import com.soywiz.klock.*
-import com.soywiz.korge.android.*
-import com.soywiz.korge.view.*
+import android.annotation.SuppressLint
+import android.os.VibrationEffect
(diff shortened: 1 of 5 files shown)


soywiz:
Looks nice! Thanks :)

soywiz:
Regarding to the module issue Im not fully sure why is that happening. Ill check if I can reproduce it

PR #77, opened 2020-01-27
Update skins for UIViews

RezMike:
Add additional skins for different UIViews.


soywiz:
What's the status of this? Need reviewing, or just more work?

RezMike:
It's just a draft now. I don't have enough time these days to continue to work on the PR. When I have time, I'll add more significant changes. When the PR is done, I'll request code review.

soywiz:
Sorry for the delay.

So you are generating two different skins for checked and unchecked?
I would rater generate a single image with all the stuff there so people wanting to create their own skin can just load one image.
Also not sure 9-patch would work fine for the checkmark image. Why not place just the checkmark symbol and place it over the button? You can add one checkmark for enabled and other for disabled in the original skin image (instead of having two separate images).

Also consider minimizing the PNG image before encoding it to base64 using ImageOptim: https://imageoptim.com/mac

checked
unchecked

RezMike:
About a single image:
Good thought, I agree with you, so I'll do it.

About 9-patch image and checkmark symbol:
Now I'm thinking about creating complex skins or IconButton class, where an icon or a symbol (like a checkmark here) can be added over the default skin with some customizable padding. What do you think about this solution?

soywiz:
Nice. Okay to the symbol+padding part 👍

RezMike:
Current results:
1. Implement IconButton and IconSkin.
2. Add check, up, down, left and right icons (as default elements).
3. Update UICheckBox, UIScrollBar, UIScrollableArea, UIComboBox. Now they use icons instead of text.
4. Add complex skins - ScrollBarSkin and ComboBoxSkin (also as default elements).

RezMike:
Here are png and psd files (in zip archive) with current DefaultUISkin image (for future updates).
default_ui_skin
[default_ui_skin.zip](https://github.com/korlibs/korge/files/4243738/default_ui_skin.zip)

RezMike:
And that's how ui elements look like now.

image

TobseF:
Before commit... just another software for png resizing for Mac & Windows:
https://nukesaq88.github.io/Pngyu/

Should we provide a default theme, or should we make the skin optional? Of course the theme is small, but I'am always surprised to find the `korge-ui.psd` and `korge-ui.png` in my target folder. - Even if I don't use the UI at all. Ind If I provide my own, I can't exclude the build in. 

Maybe we can auto convert LibGdx Skins, an we will get a big theme collection in one rush.
https://ray3k.wordpress.com/artwork/

RezMike:
There are no psd or png files in the ui package. The single png image used there is optimized, encoded to a base64 string and will be decoded lazily. All default skins are lazy-evaluated, too. So if you don't want to use default elements (skins, icons, etc.), there will be no overhead except a few files with code.

I added the png and psd files here just for a case if some other developer would like to change default ui skin and icons.

soywiz:
Thanks! We could consider supporting libgdx skins, but in a separate PR

PR #94, opened 2020-03-03
Add reified typed register to Bus

TobseF:
**New:** Adds a reified version of the register method to get a nicer API.
So instead of writing:
``` kotlin
bus.register(SomeEvent::class){ someEventListener() }
```
We can provide the type in a generic way without the `::class`.
``` kotlin
bus.register<SomeEvent> { someEventListener(it) }
```

**Refactoring:** Removes the unsed `InjectedHandler`


soywiz:
Thanks!

PR #106, opened 2020-03-17
Add additional onCancel callback to Animator

RezMike:
Add `onCancel` callback to Animator in order to run some actions when animation is cancelled.
Also wrap execution of Parallel node in try-catch block with CancellationException.


soywiz:
Thanks!

PR #117, opened 2020-03-22
 Create filtering field inside ShaderFilter

### LINKED ISSUE #107: ShaderFilter smooths view's texture
Shader filter smooths texture without taking into account if the view has **_filtering = false_**, it's a headache for pixel art gamedevs. The problem is in the below line of code:

https://github.com/korlibs/korge/blob/e1efe95e30a5d479ae35c7aac581c5149f278a86/korge/src/commonMain/kotlin/com/soywiz/korge/view/filter/ShaderFilter.kt#L94

The solution could be to modify the render method declaration inside **_Filter_** interface to include **_filtering: Boolean_** parameter.

https://github.com/korlibs/korge/blob/e1efe95e30a5d479ae35c7aac581c5149f278a86/korge/src/commonMain/kotlin/com/soywiz/korge/view/filter/Filter.kt#L45-L54

The problem with that approach is that **_View_** class doesn't have a filtering field, only inherited views like _**Text**_ and _**Image**_ (smooth field) have, so the **_renderFiltered(...)_** method inside **_View_** doesn't have a way to know whether or not the inherited class has filtering field. It might be solved creating an interface or **_filtering_** field inside **_View_** class(I came from Java world, maybe with a Kotlin feature you can get a cleaner solution).

i-d-c-p:
To allow devs to set it false when making pixel art games. Tries to fix #107

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/filter/ShaderFilter.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/filter/ShaderFilter.kt
@@ -29,6 +29,8 @@ abstract class ShaderFilter : Filter {
         fun Program.Builder.tex(coords: Operand) = texture2D(DefaultShaders.u_Tex, coords / u_TextureSize)
     }
 
+    var filtering = true
+
     private val textureSizeHolder = FloatArray(2)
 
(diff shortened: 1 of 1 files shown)


soywiz:
Thanks!

PR #126, opened 2020-04-28
Add TouchEvent handling so TouchComponent is called

soywiz:


RezMike:
@soywiz what about this PR? There are conflicts that must be resolved.

soywiz:
It seems that includes two unrelated commits. Maybe I missed to switch to the right branch. Let me check

PR #131, opened 2020-05-01
Fixes #113 compute right bounds for Graphics view

### LINKED ISSUE #113: Implement correct width and height for Graphics
A Graphics instance always has width and height equal to 1.0 despite the fact that it may contain elements (shapes) with width and height more than 1.0.

soywiz:


RezMike:
@soywiz, conceptual question. What position should have a `Circle(radius = 32)` object inside a container by default?

RezMike:
Tests fail because of incorrect position of `Circle(32)` view. I changed the center of a Circle from `(0, 0)` to `(radius, radius)`, but now it fails because of a small calculation error when drawing arc. Here are points of the arc of the Circle:
image
So now the Circle's size is not `(64.0, 64.0)`, but `(64.0, 32.00000000000001)`.

soywiz:
It should be at `0,0` with bounds `[·32,-32]-[+32,+32]`(width=64, height=64)

RezMike:
OK, I can add a commit with position fix, so `Circle(radius = 32.0)` will be located inside its parent container between [0; 64] x-axis and [0; 64] y-axis. But I don't no what to do with that size calculation error.

soywiz:
Sorry I assigned you for reviewing since it tried to fix an issue you reported. I assigned you before the test was failing and didn't know it was going to fail. Let me check it if I can find the issue

RezMike:
It's ok, since you didn't anything here in the last 6 days, I decided to check what's wrong here

korge/src/commonMain/kotlin/com/soywiz/korge/view/Circle.kt
```diff
@@ -8,37 +9,39 @@ import com.soywiz.korma.geom.vector.*
...
-
     init {
         updateGraphics()
     }
 
     private fun updateGraphics() {
         clear()
         fill(color) {
-            circle(0.0, 0.0, radius)
+            circle(radius, radius, radius)
```
soywiz, line 44: I don't think that's the right position. Unless the circle/ellipse was provided by a bounds, I think that by the principle of least surprise it should be at `[0, 0]`. But indeeed if this failed, somewhere else there might be a bug

RezMike:
The original error (why tests failed) was because of `Circle(32)` being positioned at [-32; -32] instead of correct [0; 0]. I fixed it in the last commit. But now the error is caused by the size calculation error (I described it above).

korge/src/commonMain/kotlin/com/soywiz/korge/view/Circle.kt
RezMike: Some time ago I tried to use `Circle` view, but found this error - wrong position of `Circle`. I needed to call `position(radius, radius)` by myself in order to make it fully visible. So I think its position should be like now.

RezMike: When drawing `circle` inside your own `Graphics` view, positioning circle in [0; 0] is correct. But as soon as you use `Circle` view, this view's top left corner should be positioned in [0; 0], not its center.

soywiz: I see your point. Since Graphics extends `RectBase` it seems it have `anchorX` and `anchorY` and extension methods:

```
inline fun <T : RectBase> T.anchor(ax: Number, ay: Number): T =
	this.apply { this.anchorX = ax.toDouble() }.apply { this.anchorY = ay.toDouble() }

fun <T : RectBase> T.center(): T = anchor(0.5, 0.5)
val <T : RectBase> T.centered: T get() = anchor(0.5, 0.5)
```

So I guess that's fine. People can use `.centered` if required.

korge/src/commonMain/kotlin/com/soywiz/korge/view/Circle.kt
```diff
@@ -8,37 +9,39 @@ import com.soywiz.korma.geom.vector.*
...
-        }
+    var radius: Double by uiObservable(radius) { updateGraphics() }
+    /** Color of the circle */
+    var color: RGBA by uiObservable(color) { updateGraphics() }
 
     override val bwidth get() = radius * 2
     override val bheight get() = radius * 2
 
-    /** Color of the circle */
-    // @TODO: We could paint a white circle, and then tint it using the [colorMul] property
```
RezMike, line 28: @soywiz, since `Circle` view is still a `Graphics` and a `Container`, someone can add other views inside it. If we use `colorMul` to change the color of the circle, I suppose, we will add incorrect behavior for that case (`colorMul` will be applied to other views inside `Circle`).

soywiz: Makes sense.

https://github.com/korlibs/korge/pull/131/commits/39f2d77209cf70110a3839c3a488d9e7efee0ebc

soywiz:
We have to rerun the tests once this: https://github.com/korlibs/korim/runs/650554368 is merged

soywiz:
Fucky my life. Trying to fix this it seems that I have removed all the commits doing a push force. Do you have locally your changes?

RezMike:
Yep, here there are.

PR #136, opened 2020-05-06
Implement SwipeComponent

RezMike:
Add SwipeComponent (actually, just `onSwipe` function) that lets you define action when swipes are performed. It lets you specify an optional threshold, after which a swipe should be triggered, and an optional direction for swipes (all directions by default).
Also deprecate DragComponent with indication of using `onMouseDrag` instead.

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/input/SwipeComponent.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/input/SwipeComponent.kt
@@ -8,7 +8,13 @@ data class SwipeInfo(
     var dx: Double = 0.0,
     var dy: Double = 0.0,
     var direction: SwipeDirection
-)
+) {
+    fun setTo(dx: Double, dy: Double, direction: SwipeDirection) = this.apply {
+        this.dx = dx
+        this.dy = dy
+        this.direction = direction
+    }
+}
 
 enum class SwipeDirection {
     LEFT, RIGHT, TOP, BOTTOM
@@ -24,7 +30,7 @@ enum class SwipeDirection {
  * @property callback a callback to execute
  */
 fun <T : View> T.onSwipe(
-    threshold: kotlin.Number = -1,
+    threshold: Double = -1.0,
     direction: SwipeDirection? = null,
     callback: suspend Views.(SwipeInfo) -> Unit
 ): T {
@@ -37,9 +43,10 @@ fun <T : View> T.onSwipe(
     var movedRight = false
     var movedTop = false
     var movedBottom = false
-    val thr = threshold.toDouble()
+
     val view = this
     val mousePos = Point()
+    val swipeInfo = SwipeInfo(0.0, 0.0, SwipeDirection.TOP)
 
     fun views() = view.stage!!.views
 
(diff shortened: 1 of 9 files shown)


korge/src/commonMain/kotlin/com/soywiz/korge/input/SwipeComponent.kt
```diff
@@ -0,0 +1,139 @@
+package com.soywiz.korge.input
+
+import com.soywiz.korge.view.*
+import com.soywiz.korma.geom.*
+import kotlin.math.*
+
+data class SwipeInfo(
+    var dx: Double = 0.0,
```
soywiz, line 8: This is mutable, but then you construct for every event.
To be consistent with the rest of the events, this might be mutable indeed but you should construct an instance for the `onSwipe` block and trigger that instance these fields updated

korge/src/commonMain/kotlin/com/soywiz/korge/input/SwipeComponent.kt
```diff
@@ -0,0 +1,139 @@
...
+        if (mousePos.x > cx) movedRight = true
+        if (mousePos.y < cy) movedTop = true
+        if (mousePos.y > cy) movedBottom = true
+        cx = mousePos.x
+        cy = mousePos.y
+    }
+
+    suspend fun triggerEvent(direction: SwipeDirection) {
+        register = false
+        views().callback(SwipeInfo(cx - sx, cy - sy, direction))
```
soywiz, line 65: To be consistent you should construct a dummy SwipeInfo instance and store it locally, then update it and send that same instance.
It is done like that to avoid allocations when using events.

korge/src/commonMain/kotlin/com/soywiz/korge/input/SwipeComponent.kt
```diff
@@ -0,0 +1,139 @@
...
+ * This methods lets you specify a [callback] to execute when swipe event is triggered.
+ * @property threshold a threshold for dx or dy after which a swipe event should be triggered.
+ * Negative [threshold] means a swipe event should be triggered only after [onUp] mouse event.
+ * [threshold] is set to -1 by default.
+ * @property direction a direction for which a swipe event should be triggered.
+ * [direction] is set to null by default, which means a swipe event should be triggered for any direction.
+ * @property callback a callback to execute
+ */
+fun <T : View> T.onSwipe(
+    threshold: kotlin.Number = -1,
```
soywiz, line 27: Not a big deal. I usually create a separate `inline fun` function when using `Number` and the non inline versions use Double to try to avoid boxings. Thought it is not a big deal here since this is probably not going to be called too often, but once per scene maybe

korge/src/commonMain/kotlin/com/soywiz/korge/input/SwipeComponent.kt
```diff
@@ -0,0 +1,139 @@
...
+            SwipeDirection.RIGHT -> if (cx - sx > thr && !movedLeft) triggerEvent(direction)
+            SwipeDirection.TOP -> if (sy - cy > thr && !movedBottom) triggerEvent(direction)
+            SwipeDirection.BOTTOM -> if (cy - sy > thr && !movedTop) triggerEvent(direction)
+            null -> when {
+                sx - cx > thr && !movedRight -> triggerEvent(SwipeDirection.LEFT)
+                cx - sx > thr && !movedLeft -> triggerEvent(SwipeDirection.RIGHT)
+                sy - cy > thr && !movedBottom -> triggerEvent(SwipeDirection.TOP)
+                cy - sy > thr && !movedTop -> triggerEvent(SwipeDirection.BOTTOM)
+            }
+        }
```
soywiz, line 81: There is duplicated code. What if you remove the `when (direction)` completely, then defined if trigger or not based on th provided direction at the `triggerEvent` method? Like `if (providedMethod != null && providedMethod != triggerDirection) return`

korge/src/commonMain/kotlin/com/soywiz/korge/input/SwipeComponent.kt
```diff
@@ -0,0 +1,139 @@
...
+                    triggerEvent(SwipeDirection.RIGHT)
+                }
+            } else {
+                if (cy < sy && !movedBottom) {
+                    triggerEvent(SwipeDirection.TOP)
+                } else if (cy > sy && !movedTop) {
+                    triggerEvent(SwipeDirection.BOTTOM)
+                }
+            }
+        }
```
soywiz, line 106: Same here, if `triggerEvent` does the check, you would de-duplicate a lot of code here too. If you prefer, you can create a separate method called `triggerEventIfRequired` or `maybeTriggerEvent` to do that check

soywiz:
Also it seems that it is failing on linux because master was broken when you forked/updated it. I have fixed master already.

PR #148, opened 2020-05-11
Fixes #112 : initial size of the Stage was wrong

### LINKED ISSUE #112: Incorrect root.width and root.height in Korge block
When I comment the first line, the root.width is much greater than the actual width of the window, I need to use root.height to get the width value. But with the first line, the root.width and root.height are correct.

image

soywiz:
This is a temporal workaround. This delays the execution of the main entrypoint just before the first frame. But we can and should do better later by simplifying some stuff at the korgw level.


RezMike:
It fixes my issue, but it causes TimeoutCancellationException in tests that use `viewsTest` function. I suppose, the reason is that there are no RenderEvents triggered in tests. Anyway, it should be fixed, so I can't merge it now.

Issue #125, opened 2020-04-26
Add Sprite class


soywiz:
A class to create simple animation based on several images from an atlas or not.

soywiz:
Reference/proposal here: https://github.com/emign/animationcontrollerexample/commit/0d82b4466c9219a86547c5310121a026f3e78ec1

emign:
Another propasal
https://github.com/korlibs/korge/pull/149

PR #149, opened 2020-05-12
Added SpriteAnimation.kt

emign:
For a sample how to use it see PR from korge-samples:

https://github.com/korlibs/korge-samples/pull/13

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/Sprite.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/Sprite.kt
@@ -1,69 +1,98 @@
 package me.emig.engineEmi.graphics.animationen
 
 import com.soywiz.klock.TimeSpan
-import com.soywiz.klock.measureTime
 import com.soywiz.klock.milliseconds
 import com.soywiz.korge.view.Image
-import com.soywiz.korio.async.delay
+import com.soywiz.korge.view.addUpdater
+import com.soywiz.korim.bitmap.Bitmap
+import com.soywiz.korim.bitmap.Bitmaps
+import com.soywiz.korim.bitmap.BmpSlice
+import com.soywiz.korio.async.Signal
 
-class Sprite(vararg var animations: SpriteAnimation) : Image(animations[0].firstSprite) {
-    var stop = false
-    var currentAnimation : SpriteAnimation = animations[0]
-
-    suspend fun playAnimationForDuration(duration: TimeSpan, id : String = "", spriteDisplayTime: TimeSpan = currentAnimation.frameTime) {
-        renewCurrentAnimation(id)
-        stop = false
-        if (duration > 0.milliseconds) {
-            var timeCounter = duration
-            while (timeCounter > 0.milliseconds && !stop ) {
-                val measuredTime = measureTime {
-                    nextSprite()
-                    delay(spriteDisplayTime)
+class Sprite(bitmap : Bitmap) : Image(bitmap) {
+    constructor(bmpSlice : BmpSlice) : this(bmpSlice.bmp)
+    constructor(initialAnimation : SpriteAnimation) : this(initialAnimation.firstSprite){
+        currentAnimation = initialAnimation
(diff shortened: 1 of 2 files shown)


soywiz:
I added a few comments. The idea looks great and looks useful. What do you think about the part of splitting the class in two so we can reuse animations? A view + a reusable sprite/animation definition

emign:
Give me some time to implement these. 
I think your suggestion for the split in two classes is the better approach. I was somehow too fixed on the Animation itself to realize the bigger picture.

soywiz:
Sure. No rush! Thanks for the contribution

korge/src/commonMain/kotlin/com/soywiz/korge/view/Sprite.kt
```diff
@@ -0,0 +1,69 @@
...
+
+import com.soywiz.klock.TimeSpan
+import com.soywiz.klock.measureTime
+import com.soywiz.klock.milliseconds
+import com.soywiz.korge.view.Image
+import com.soywiz.korio.async.delay
+
+class Sprite(vararg var animations: SpriteAnimation) : Image(animations[0].firstSprite) {
+    var stop = false
+    var currentAnimation : SpriteAnimation = animations[0]
```
soywiz, line 11: Some people might call `Sprite()` without arguments and get an exception here. Thinking on it, maybe we can deal 
 with the complexity of a set of animations in other place.
For example, instead of passing the id for playAnimation, passing the `SpriteAnimation` itself? With an initial one in the constructor? Each approach have its advantages and disadvantages. But I guess people could still have an animation collection or something that could be a separate class and then do something like (or just in plain local variables): `playAnimationForDuration(animation = animations["standing"])`. The Sprite class code is simplified (we only have to deal with one animation, and we know for sure it exists).
`animations` in the sample could be a simple map or a class like `SpriteAnimations` or `SpriteAnimationsCollection` handling the map.

Would be nice to know the use-cases of people. Maybe people want to just enable an animation looping forever. Maybe we can do something like (the play method is not suspend and uses the updater, thus reducing allocations and make it more stable in terms of performance/microstuttering):

```
val completedAnimation = Signal<Unit>()

init {
addUpdater { ms ->
   if (running) {
        // here take into account number of loops or something if wanted? I think there is a Times class that supports infinite loops and things like that
        // You can also trigger a signal when a loop of the animation completes so other methods can subscribe?
        // ... 
   }
}
}
```

With this approach we can do something like (not checked and incomplete):
* `fun play(animation) { this.animation = animation; this.frame = 0; running = true }` 
* `fun stop { running = false }`
* `suspend fun playAnimationForDuration() { running = true; delay(duration)  }`
* `suspend fun playAnimationTimes() { running = true; while (times.notZero) { completedAnimation.await()  ; times = times.prev } }`

It's late, so not too fresh and maybe I missed stuff. What do you think?

korge/src/commonMain/kotlin/com/soywiz/korge/view/Sprite.kt
```diff
@@ -1,69 +1,98 @@
...
-        stop = false
-        val cycleCount = times*currentAnimation.spriteStackSize
-        var cycles = 0
-        while (cycles < cycleCount && !stop) {
-            nextSprite()
-            delay(spriteDisplayTime)
-            cycles++
-        }
+    fun playAnimation(spriteAnimation: SpriteAnimation) {
+        updateCurrentAnimation(
```
emign, line 51: this could be reworked as an expression call.

soywiz:
Awesome. Looks super nice. I will review later with more detail.
Can you add builder methods to be consistent with the rest of the framework? Builder constructing and adding the view like this one:

https://github.com/korlibs/korge/blob/ad589ddd1e6248199d2179fb9ea333962507cc57/korge/src/commonMain/kotlin/com/soywiz/korge/view/SolidRect.kt#L6-L8

Since you have several constructor signatures, would be nice to support them in the constructors too, or at least the most common one

soywiz:
Fixes #125

soywiz:
Thanks!

Issue #129, opened 2020-04-30
Layout support


soywiz:
Bring back korui or evaluate other options/improve layouts:

https://github.com/korlibs/korgw/tree/82860b68faae2ee5732706d7eef52049a100b6b8/%40old/korui

soywiz:
Partially included on KorGE 2.0

PR #152, opened 2020-05-14
Add relative positioning for views

RezMike:
Add special functions for views that let you position a view relative to other view (align this view's left/right side to other view's left/right side and this view's top/bottom side to other view's top/bottom side). Also add tests for centering and positioning functions and add some fixes.

This may cover many cases for layouts (#129). Use of these functions and simple containers (for grouping views) will cover (almost) all cases for layouts.


RezMike:
Padding and spacing can be implemented by a simple code line like `x += 10.0` or `y -= 20.0` (for padding/spacing 10px horizontally and 20px vertically). I don't think we need to add additional complex functions to implement this simple padding/spacing behavior (and I suppose, there will many functions for that and they will be much less intuitive).

RezMike:
About `Anchor` - I haven't used it yet. But your suggestion is interesting. It probably will be useful if we add support for `Anchor` class, something like `alignTo(other: View, anchor: Anchor`).
When a developer uses `leftToRight` or other functions from this PR, he won't need to use additional functions for padding. But when he uses `Anchor` for aligning, he probably will need this additional padding value.
So, it's a good suggestion. I'll need to try `Anchor` and to see what we can do with it.

RezMike:
I decided to rename functions (add `align` word: rename `leftToLeft` to `alignLeftToLeft`). This way it'll be more clear and easier for new developers to find these functions.

soywiz:
About incrementing or decrementing the x or y for padding: yes, you can do that. I just ser that more as a matter of semantics, communication and reading `func(padding = 15)` or `x += 15` whats that 15 for? It increments the x or decrements the y, but what means? You might need to think more about it when reading. Also depending on the alignment direction positive and negative could mean adding or subtracting the padding.

RezMike:
Do you suggest to add functions like `view.paddingLeft(20.0)`?
Hm, actually, it could be quite good with chaining calls:
```
solidRect(20, 20, Colors.WHITE)
    .positionY(30)
    .alignLeftToRight(otherRect)
    .paddingLeft(20)
```

RezMike:
Ah, you probably suggest that:
```
solidRect(20, 20, Colors.WHITE)
    .positionY(30)
    .alignLeftToRight(otherRect, 20)
```

soywiz:
Yes, i suggested the last one. But I would use it with a named parameter to give meaning to that number:

```
solidRect(20, 20, Colors.WHITE)
    .positionY(30)
    .alignLeftToRight(otherRect, padding = 20)
```

RezMike:
Well, I don't know if it's clear what this code should mean.
```
val rect = solidRect(...).position(100, 100)
val other = solidRect(...).alignRightToRight(rect, 20)
```
If `rect` is under `other`, what position should they have relative to each other?

soywiz:
For me, if you are placing it inside the other object the padding would go further inside the other shape, moving away from the edge. So positive padding would be -x.
If you think this is controversial we can omit the padding for now, and rethink it later. I think this could be merged now as it is adding value already.

RezMike:
When you write `other.apply { x += 20 }`, you know that `other` moves right by 20px.

emign:
For me who didn’t follow any of these commits on code level : I find a name like `alignrighttoright` somewhat difficult to understand

RezMike:
Well, I think, we can add `padding` in this PR, but first I want to draw an image showing the correct behaviour

RezMike:
@emign do you have any suggetions about renaming these functions? Any help is appreciated.

RezMike:
**Horizontal aligning:**
image

RezMike:
**Vertical aligning:**
image

RezMike:
@soywiz, is this the correct behavior of `align` and `padding`?

RezMike:
Sorry, the first image was incorrect, I updated it.

emign:
maybe:
```
rect1.toLeftOf(rect1)

rect1.toTopOf(rect2, +10)
```

RezMike:
About the first code sample: how do you know which side you're aligning to left/top?
About the second code sample: it looks less intuitive and too verbose.

emign:
First one is with default anchor = left

Well thats a matter of taste I would say. You are more into the code you already wrote. I just saw the calls of `alignLeftToRight` and did not understand what they do. I was more thinking of layout options for RTL layouts. 
For me `.alignToLeftOf()` is much more intuitive.

soywiz:
First of all, don't worry too much about this, we can evolve the API if required. Either deprecating, removing, renaming or whatever.

I really don't know what would be more clear. Ideally for me the layouting would be done with a designer and not by code whenever possible for non-simple. games. We don't have tools for this yet. And would be nice to have something at intellij to layout or position objects: https://github.com/korlibs/korge-intellij-plugin/issues/7

For complex games I used Adobe Animate for layouts, positioning and animations, then just decorated those views by code and did things like: "move this view to the position to the other view; I don't care what position is that one". "Or just play this animation in this object".

Checking the Affinity Designer:

Screenshot 2020-05-14 at 21 02 01

They just specify "Align Left", "Align Right" or whatever. No two directions. But maybe you are covering additional relative layouting right?
I think that the images you done are nice. We should put them in the documentation 👍

I'm not too used to this to provide further valuable feedback here. Maybe we can ask someone else in slack to provide feedback about their name and untie?

RezMike:
@emign but when you write `view.alignToLeftOf(otherView)`, you don't know exactly, what side of `view` you're aligning (left or right). If it's set to left by default - you don't know that unless you saw the source code or docs. If we make left side by default, what about people who used to write from right to left (for them the intuitive default value would be right, not left).

So this is the same situation for you:
> You are more into the code you already wrote.

RezMike:
@soywiz in graphics editors there are only three options maybe because when you select many objects (like three or more) it's not clear how they should be aligned when you try to align them `rightToLeft` (should they go one after another? then in what order?)

> But maybe you are covering additional relative layouting right?
>...
> I'm not too used to this to provide further valuable feedback here.

Well, I'm an android developer and I use `ConstraintLayout` in Android pretty often because it's a very convenient and useful way to layout views in Android. So I decided to add a similar approach in korge. You can check positioning and other stuff of `ConstraintLayout` in [this official guide](https://developer.android.com/reference/androidx/constraintlayout/widget/ConstraintLayout?hl=sr#developer-guide).

>I think that the images you done are nice. We should put them in the documentation 👍

Thanks! Yes, you can add them in the documentation, it'd be very useful.

soywiz:
When blocked by not sure how to do something, I usually ask other people, or just do something in a way that can be changed later. But in the end if there are similar options elsewhere, it is a good thing to follow other conventions so people when facing this. https://en.wikipedia.org/wiki/Principle_of_least_astonishment
So for me, at least if Android use those names, it is popular enough to be taken into account. If those names are used elsewhere would be even a better objective point supporting that proposal. So +1 from me for Android, +2 if you can find other framework using those same naming conventions.

RezMike:
@soywiz, here are updated images for the documentation:

**Horizontal aligning:**
image

**Vertical aligning:**
image

RezMike:
@soywiz, @emign, thanks for your comments!

PR #159, opened 2020-05-20
Added View.sceneContainer, View.scene, View.views and View.onAttachToStage

soywiz:
https://soywiz.slack.com/archives/C03B2H44G/p1589966531017000

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/View.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/View.kt
@@ -108,6 +108,7 @@ abstract class View : Renderable, Extra by Extra.Mixin(), EventDispatcher by Eve
 	var speed: Double = 1.0
     /** Parent [Container] of [this] View if any, or null */
 	var parent: Container? = null; internal set
+
     /** Optional name of this view */
 	var name: String? = null
     /** The [BlendMode] used for this view [BlendMode.INHERIT] will use the ancestors [blendMode]s */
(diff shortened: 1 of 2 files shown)


korge/src/commonMain/kotlin/com/soywiz/korge/view/View.kt
```diff
@@ -978,6 +979,28 @@ class ViewTransform(var view: View) {
...
+    // @TODO: On KorGE 2.0 we should have components for subscribing to attaching and detaching from the stage.
+    // @TODO: We would keep a list of views with components attached to the Stage to avoid iterating the tree each time
+    addComponent(object : UpdateComponent {
+        override val view: View = this@onAttachToStage
+        override fun update(ms: Double) {
+            val views = view.views
+            if (views != null) {
+                removeFromView()
+                block(views, view)
+            }
```
RezMike, line 999: What if this view is being removed from container and added to it later? This component will work only for the first attachment to the stage.

soywiz:
Two things:
* @RezMike you are right. I tried to implement it directly in the view, but the implementation let to a lot of overhead to all the parents and childrens. So I decided to do it like that.
* This functionality already existed if you want to accept the overhead. I just didn't remember: https://github.com/korlibs/korge/blob/9516dd31c421e7f89c66ba183518ae342475273b/korge/src/commonMain/kotlin/com/soywiz/korge/component/StageComponent.kt

I'm closing this PR. Let's see if in KorGE 2.0 we can make that with less overhead.

PR #158, opened 2020-05-20
Feature/korge.optimizations

soywiz:
Profiling 10K sprite sample. I did these optimizations.

Plus one typo in the sample itself that affected the performance a lot:
https://github.com/korlibs/korge-samples/commit/f3149874791cc47b7399ef1f896288b8f61972ae

PR #165, opened 2020-05-22
SpriteAnimation Bugfixes and optimizations + tests

### LINKED ISSUE #157: Spriteanimation ignores times parameter when spriteTimeDisplay is set.
```
view(1, spriteDisplayTime = 1.seconds)
```
this will play more than 1 time
```
quest.playAnimation(1)
```
this will play correctly one time

emign:
- Triggers working correctly
- Parameter combinations work (even weird ones)
- Played duration gets measured with higher precision
Fixes #157


korge/src/commonTest/kotlin/com/soywiz/korge/view/ViewsTest.kt
```diff
@@ -247,4 +251,98 @@ class ViewsTest : ViewsForTesting() {
         assertEquals(rect.x, 0.0)
         assertEquals(rect.y, -5.0)
     }
+
+    @Test
+    fun testSpriteAnimationPlay1Times() = viewsTest {
```
RezMike, line 256: Maybe you could extract animation tests in a new test class (`AnimationTest.kt`, for example)? The `ViewsTest` class is already looks overloaded, and it seems like quite a lot new tests only for animation.

emign: Thx. I can do that.

korge/src/commonMain/kotlin/com/soywiz/korge/view/Sprite.kt
```diff
@@ -39,68 +39,68 @@ inline fun Container.sprite(
...
-            return _onAnimationStarted!!
-        }
-
-    var spriteDisplayTime : TimeSpan = 50.ms
+    var spriteDisplayTime: TimeSpan = 50.milliseconds
     private var animationLooped = false
-    private var lastAnimationFrameTime  : TimeSpan = 0.ms
-    private var animationRequestedDuration : TimeSpan = 0.ms
+    private var lastAnimationFrameTime: TimeSpan = 0.milliseconds
+    private var animationRemainingDuration: TimeSpan = 0.milliseconds
```
soywiz, line 84: This optimization was lost: https://github.com/korlibs/korge/pull/158/commits/e147ae43ff37bf9e035946cceeb2b39f2b83dc9b

korge/src/commonMain/kotlin/com/soywiz/korge/view/Sprite.kt
```diff
@@ -39,68 +39,68 @@ inline fun Container.sprite(
...
-            if (_onAnimationStopped == null) _onAnimationStopped = Signal()
-            return _onAnimationStopped!!
-        }
-
-    val onAnimationStarted: Signal<SpriteAnimation>
-        get() {
-            if (_onAnimationStarted == null) _onAnimationStarted = Signal()
-            return _onAnimationStarted!!
-        }
-
```
soywiz, line 91: This optimization was lost: https://github.com/korlibs/korge/pull/158/commits/5aab8def22baa39b4b51dc824aaa2c06419645d0

Some Sprite won't have subscribers, and you are creating several objects and doing iterations when dispatching them even with no subscribers. The 10K Sprites sample benefitied from this optimization

emign: I will reintroduce the changes. Have they been Public already ?

soywiz: They are on master already but not published a version yet. You probably forked the repo before that was merged

emign: Seems so. Well. Will be fixed :)

korge/src/commonTest/kotlin/com/soywiz/korge/view/ViewsTest.kt
```diff
@@ -247,4 +251,98 @@ class ViewsTest : ViewsForTesting() {
...
+
+    @Test
+    fun testSpriteAnimationReversedStartFrame() = viewsTest {
+        val countDownSpriteMap = resourcesVfs["countDown.png"].readBitmap().toBMP32()
+        val countDownAnimation = SpriteAnimation(countDownSpriteMap,24,36,0,0,10,1,0,0)
+        val countDownSprite = Sprite(countDownAnimation)
+
+        val digits = Array<Bitmap>(10){
+            countDownSpriteMap.extract(it*24, 0,24,36)
+        }
```
soywiz, line 328: If you move these tests to its own class, you can also move these locals into `lateinit var` properties, then use a `@Before` or a plain `suspend fun` to populate them, since seems to be the same for all the tests, right?

Issue #167, opened 2020-05-23
Optimize for Kotlin/Native


soywiz:
* Try to find places where we are using `open fun`, `open val` or `open var` where we can use a final one (avoid dinamic dispatch) and in some cases devirtualization doesn't work which forces a openmethod lookup
* Try to avoid some interfaces when possible
* Try to reduce usages for `is` and `as` (there is a patch to optimize this on Kotlin/JS provided by KorGE-plugin and K/N has improved it, but that might prevent some devirtialization)
* Suggest to the Kotlin/Native team to allow to disable Mutability checks to remove that overhead in release builds or with another flag

PR #168, opened 2020-05-23
Feature/native.optimizations

soywiz:
Some work on #167


soywiz:
This makes Kotlin/Native and JS to handle more or less well the 10K sprite sample

Issue #216, opened 2020-06-15
Sprite / SpriteAnimation playing ends with the first frame instead of the last one


soywiz:
https://kotlinlang.slack.com/archives/CJEF0LB6Y/p1592177764247500

emign:
We don’t have a parameter defining an `endFrame`. It’s a feature we should add though

PR #224, opened 2020-06-16
Add end frame for sprite animations

emign:
Adds #216


korge/src/commonMain/kotlin/com/soywiz/korge/view/Sprite.kt
```diff
@@ -235,11 +243,17 @@ open class Sprite(
         animationType = type
         animationRequested = true
         currentAnimation?.let {
-            this.animationNumberOfFramesRequested = if (!looped) animationCyclesRequested * it.spriteStackSize else it.spriteStackSize
+            this.animationNumberOfFramesRequested = when {
```
emign, line 246: That's what came out of my mind. I feel, this could be handled more elegantly though. Maybe you have ideas

RezMike:
Fixes #216

PR #229, opened 2020-06-18
Fix KERNEL_SHARPEN in Convolute3Filter

RezMike:


korge/src/commonMain/kotlin/com/soywiz/korge/view/filter/Convolute3Filter.kt
```diff
@@ -21,37 +21,37 @@ class Convolute3Filter(
...
         )
 
         /** A Sharpen Kernel. This [Matrix3D] can be used as [kernel] for [Convolute3Filter] */
         val KERNEL_SHARPEN: Matrix3D = Matrix3D.fromRows3x3(
-            0f, -1f, 0f,
-            -1f, +4f, -1f,
-            0f, -1f, 0f
+            -1.0, -1.0, -1.0,
+            -1.0, +9.0, -1.0,
+            -1.0, -1.0, -1.0
```
soywiz, line 54: Where did you get those numbers?
https://en.wikipedia.org/wiki/Kernel_(image_processing)

soywiz: 4 is wrong, but should be:
```
0, -1, 0,
-1, +5, -1,
-1, 0, -1
```

korge/src/commonMain/kotlin/com/soywiz/korge/view/filter/Convolute3Filter.kt
```diff
@@ -21,37 +21,37 @@ class Convolute3Filter(
...
         /** A Gaussian Blur Kernel. This [Matrix3D] can be used as [kernel] for [Convolute3Filter] */
         val KERNEL_GAUSSIAN_BLUR: Matrix3D = Matrix3D.fromRows3x3(
-            1f, 2f, 1f,
-            2f, 4f, 2f,
-            1f, 2f, 1f
-        ) * (1f / 16f)
+            1.0, 2.0, 1.0,
+            2.0, 4.0, 2.0,
+            1.0, 2.0, 1.0
+        ) * (1.0 / 16.0)
```
soywiz, line 27: This is incidentally wrong. Matrix3D uses FloatArray as backing array.
I think we should update KorMA to support function calling using Floats in addition to doubles here: https://github.com/korlibs/korma/blob/master/korma/src/commonMain/kotlin/com/soywiz/korma/geom/Matrix3D.kt

korge/src/commonMain/kotlin/com/soywiz/korge/view/filter/Convolute3Filter.kt
RezMike: Here
http://www.roborealm.com/help/Convolution.php

RezMike: Both matrix (with +5 and +9 in the center) works, but the one with +9 in the center shows a better result (the resulting image is sharper), so I decided to write it instead of the matrix with +5.

korge/src/commonMain/kotlin/com/soywiz/korge/view/filter/Convolute3Filter.kt
RezMike: Ok, I update the PR later

RezMike: https://github.com/korlibs/korma/pull/31

Issue #231, opened 2020-06-19
Stage.scaleX and Stage.scaleY are not 1.0 on retina displays


soywiz:
This could be unexpected. Caused issues here: https://github.com/Kooostia16/projects/tree/master/spaceRaider/src

soywiz:
Maybe will be nice that this is hiddent at the AG/KorGW level?

PR #239, opened 2020-06-28
New Camera API (experimental)

RezMike:
New Camera API for KorGE 2.0. It's work in progress now, some features are not implemented yet.


RezMike:
@soywiz I need your help to implement camera rotation [here](https://github.com/korlibs/korge/pull/239/commits/2e8df81d4f228a50f4fd7be806a96120a8753939#diff-2466e2aa0008fd4fd9168996ead96586R44-R57)

korge/src/commonMain/kotlin/com/soywiz/korge/view/camera/CameraContainer.kt
```diff
@@ -0,0 +1,67 @@
...
+        val zoom = camera.zoom
+        val x = camera.x + camera.width * (zoom - 1) / zoom * camera.anchorX
+        val y = camera.y + camera.height  * (zoom - 1) / zoom * camera.anchorY
+        mat.translate(-x, -y)
+        mat.scale(
+            this.width / camera.width * zoom,
+            this.height / camera.height * zoom
+        )
+        //TODO: soywiz, implement correct rotation with anchoring
+        //mat.rotate(camera.angle)
```
soywiz, line 55: Probably you have to use `pretranslate`, `prescale` and `prerotate`. When using rotation, the 0,0 point must be on the center first.

https://gamedev.stackexchange.com/questions/16719/what-is-the-correct-order-to-multiply-scale-rotation-and-translation-matrices-f

korge/src/commonMain/kotlin/com/soywiz/korge/view/camera/CameraContainer.kt
```diff
@@ -0,0 +1,67 @@
...
+        return mat
+    }
+
+    internal fun setTo(camera: Camera) {
+        content.localMatrix = getLocalMatrixFor(camera)
+    }
+
+    internal suspend fun tweenTo(camera: Camera, time: TimeSpan, easing: Easing) {
```
soywiz, line 63: I think this would require a job cancellation here  (that would do some GC, and maybe it won't be that smooth). But you can try.

In the past I have used an `addUpdater` here that updates the current position approaching to a target destination. 
So you would have the properties of the camera copied there, then on each frame you compare the local properties with the current camera properties then uses a formula to approach more or less the properties (unless you are setting the properties). This usually allows to change the current camera and keep the movement smooth.

korge/src/commonMain/kotlin/com/soywiz/korge/util/Observable.kt
```diff
@@ -0,0 +1,17 @@
+package com.soywiz.korge.util
+
+import kotlin.reflect.*
+
+class Observable<T>(val initial: T, val before: (T) -> Unit = {}, val after: (T) -> Unit = {}) {
```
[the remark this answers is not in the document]
RezMike: Yep, kotlin delegates doesn't fit here. I need to execute some code before the variable has changed, but I don't need to prevent the change, I just need to prevent executing this code. So yes, I will propose a PR to kds with this class.

korge/src/commonMain/kotlin/com/soywiz/korge/view/camera/Camera.kt
```diff
@@ -0,0 +1,173 @@
...
+                )
+            }
+        }
+    }
+
+    fun unfollow() {
+        cancelFollowing?.cancel()
+    }
+
+    fun copy(
```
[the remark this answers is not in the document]
RezMike: Do we need them in Camera? I can't come up with a case where it would be needed. Maybe you can give an example?

korge/src/commonMain/kotlin/com/soywiz/korge/view/camera/Camera.kt
```diff
@@ -0,0 +1,173 @@
...
+
+fun Camera.size(width: Double, height: Double): Camera {
+    return setTo(width = width, height = height)
+}
+
+fun Camera.anchor(anchorX: Double, anchorY: Double): Camera {
+    return setTo(anchorX = anchorX, anchorY = anchorY)
+}
+
+suspend fun Camera.moveTo(x: Double, y: Double, time: TimeSpan, easing: Easing = Easing.LINEAR) {
```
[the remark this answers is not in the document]
RezMike: I don't know because returning camera would be useful for chained calls and I don't think these suspend functions with the effect of code postponing should be used in chained calls (a user could confuse them with the view dsl functions).

korge/src/commonMain/kotlin/com/soywiz/korge/view/camera/CameraContainer.kt
RezMike: I think I can try tweening localMatrix of the container and camera at the same time. Maybe it will work.

korge/src/commonMain/kotlin/com/soywiz/korge/util/Observable.kt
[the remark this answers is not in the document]
RezMike: I'll update this PR once the new version of kds with https://github.com/korlibs/kds/pull/17 is released

korge/src/jvmTest/kotlin/com/soywiz/korge/view/camera/CameraTest.kt
```diff
@@ -0,0 +1,87 @@
...
+    lateinit var actor: View
+    container.updateContent {
+        actor = solidRect(50.0, 50.0, Colors.RED)
+        //actor.centerOnParent()
+    }
+    delay(1.seconds)
+    cam.zoom(2.0, 1.seconds)
+    delay(1.seconds)
+
+    cam.follow(actor)
```
emign, line 82: Since `.camera=` is only available outside of the cameraViews containers scope, you are forced to declare `Views` the cam should follow outside the `cameraContainer` scope too. Atm I cannot follow a view, which is declared in the container itself

korge/src/jvmTest/kotlin/com/soywiz/korge/view/camera/CameraTest.kt
```diff
@@ -0,0 +1,87 @@
...
+        solidRect(400.0, 400.0, Colors.YELLOW)
+        solidRect(350.0, 350.0, Colors.YELLOWGREEN)
+        solidRect(300.0, 300.0, Colors.GREENYELLOW)
+        solidRect(250.0, 250.0, Colors.GREEN)
+        polygon(50.0, 7, Colors.DARKGREY).position(200.0, 100.0)
+    }
+
+    val cam = Camera(0.0, 0.0, 400.0, 400.0)
+    val cam2 = cam.copy()
+    val container = cameraContainer(400.0, 400.0, decoration = { solidRect(width, height, Colors.PINK) }) {
```
emign, line 32: CameraContainer on mac is only quarter the size it is supposed to have.
Probably related to #231 
Korge_und_korgeTemplate_–_main_kt__korgeTemplate_commonMain_

korge/src/jvmTest/kotlin/com/soywiz/korge/view/camera/CameraTest.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    fun elements() = Container().apply {
+        solidRect(400.0, 400.0, Colors.YELLOW)
+        solidRect(350.0, 350.0, Colors.YELLOWGREEN)
+        solidRect(300.0, 300.0, Colors.GREENYELLOW)
+        solidRect(250.0, 250.0, Colors.GREEN)
+        polygon(50.0, 7, Colors.DARKGREY).position(200.0, 100.0)
+    }
+
+    val cam = Camera(0.0, 0.0, 400.0, 400.0)
```
emign, line 30: The cam does not behave as I think it would.
```
val cam = Camera(.0, .0, cont.width, cont.height )
	cont.camera = cam
	cam.follow(player.sprite)
```
produces the display below (with the 2 times scaling issue on mac).
But 
```
val cam = Camera(.0, .0, cont.width*2, cont.height*2 )
	cont.camera = cam
	cam.follow(player.sprite)
```
produces an empty screen (only filled with the KorGE bgcolor)

PR #242, opened 2020-07-04
Fix/fixes.view.center on

soywiz:
Fixes centerOn functions that were assuming the same reference system, and it is not the case on the stage since local stage.x might not be 0 if the aspect ratio is different.

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/View.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/View.kt
@@ -1694,13 +1694,13 @@ fun <T : View> T.centerBetween(x1: Double, y1: Double, x2: Double, y2: Double): 
  * Chainable method returning this that sets [View.x] so that
  * [this] View is centered on the [other] View horizontally
  */
-fun <T : View> T.centerXOn(other: View): T = this.centerXBetween(other.x, other.x + other.width)
+fun <T : View> T.centerXOn(other: View): T = this.alignX(other, 0.5, true)
 
 /**
  * Chainable method returning this that sets [View.y] so that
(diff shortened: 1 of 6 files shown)


korge/src/commonMain/kotlin/com/soywiz/korge/view/View.kt
```diff
@@ -1679,22 +1679,22 @@ fun <T : View> T.centerBetween(x1: Double, y1: Double, x2: Double, y2: Double):
  * Chainable method returning this that sets [View.x] so that
  * [this] View is centered on the [other] View horizontally
  */
-fun <T : View> T.centerXOn(other: View): T = this.centerXBetween(other.x, other.x + other.width)
```
soywiz, line 1682: This only was guaranteed to work if  `this` and `other` have the same parent (and thus the same parent.globalTransform)

Some people did:

```
view.centerXOnStage()
```

And since stage.x might not be 0, specilaly on JS where the aspect ratio might be different, this caused the centered text to appear to the right since `(stage.x + stage.width) / 2` is not the center

RezMike:
Hm, for some reason these fixes give incorrect result in my 2048 game:

image

RezMike:
Well, that's because `image(...).size(10, 10)` doesn't change `width` and `height` of the image, it changes `bwidth` and `bheight` of it

RezMike:
I think it's another bug
@soywiz, could you fix it in this PR, too?

soywiz:
@RezMike thanks for reporting. I have added a few tests, and reproduced and fixed your case. Can you try it in your own code? Now it should work fine

RezMike:
Yes, now it works correctly.

PR #249, opened 2020-07-06
Fixes #238: Support tilemap rotated/reflected tiles

### LINKED ISSUE #238: Support tilemap rotated/reflected tiles

soywiz:


RezMike:
I can't say anything useful regarding correctness of the implementation. I'll need to start implementing tiledmap drawing, then I will be able to check this.
So for now it looks good to me. If I find bugs later, I'll fix them or create an issue.

soywiz:
I have already verified that works fine

PR #258, opened 2020-07-10
⚙️ Added sync with central versioning repo to DEPLOY.yml

emign:
For testing this live in the repo (KorGE) which is frequently changed.Should not affect deployment or publishing on bintray because it runs after publish-finalize task

### DIFF
--- a/.github/workflows/DEPLOY.yml
+++ b/.github/workflows/DEPLOY.yml
@@ -44,3 +44,28 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - run: ./gradlew actuallyPublishBintray
+
+  publishVersionNumbers:
+    needs: [publish-finalize]
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get repository name
+        run:  echo "::set-env name=REPOSITORY_NAME::$(echo "$GITHUB_REPOSITORY" | awk -F / '{print $2}')"
+
+      - name: Get version
+        run:  echo "::set-env name=VERSION::${GITHUB_REF/refs\/tags\/v/}"
+
+      - name: Checkout korlibs-versions repo
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.KORGEBOT_TOKEN }}
+          repository: korlibs/korlibs-versions
+
+      - name: Bump version numbers
+        run: |
+          sed -i 's/'$REPOSITORY_NAME'Version=.*/'$REPOSITORY_NAME'Version='$VERSION'/g' versions.ver
+          git config --global user.email "githubactionsbot@korge.org"
+          git config --global user.name "korgebot"
+          git add versions.ver
+          git commit -m "${{ env.REPOSITORY_NAME }} bump to ${{ env.VERSION }}" -m "Bumped ${{ env.REPOSITORY_NAME }} version number to ${{ env.VERSION }}"
+          git push


.github/workflows/DEPLOY.yml
```diff
@@ -44,3 +44,28 @@ jobs:
...
+        run:  echo "::set-env name=REPOSITORY_NAME::$(echo "$GITHUB_REPOSITORY" | awk -F / '{print $2}')"
+
+      - name: Get version
+        run:  echo "::set-env name=VERSION::${GITHUB_REF/refs\/tags\/v/}"
+
+      - name: Checkout korlibs-versions repo
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.KORGEBOT_TOKEN }}
+          repository: korlibs/korlibs-versions
```
soywiz, line 62: We should improve this adding something like this:

```
with:
    ref: my-branch
```

So we have `master` branch updating master, and `kotlin-1.4` updating the WIP KorGE 2.0

soywiz: https://stackoverflow.com/questions/58033366/how-to-get-current-branch-within-github-actions

soywiz:
```
- name: Extract branch name
  shell: bash
  run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
  id: extract_branch
```

emign: I have to think about that, because there is a added complexity. At the moment of pulling korlibs-versions the version branch of the current version might not exist yet.

emign: And another thing for understanding.
In the repos we have branches like `/releases/1.11.0` so we want to have a `1.11.0` branch in korlibs-versions?

soywiz: /releases branches won't be used anymore. Was something legacy. In normal circumstances we will release on master or on kotlin1-4. In the future if we have several versions at once, we should use the same structure on all the repos and create that branch on the kotlin-versions repo.

soywiz: I created `kotlin-1.4` branch the other day: https://github.com/korlibs/korlibs-versions/tree/kotlin-1.4

emign: So what are the conditions for committing to `korlibs-versions/kotlin-1.4` ?
a `v2` prefix?

soywiz: No. One thing is the branch, and the other the tag. I hope that from the current version you can get the tag, but also the branch containing that commit.

For example:

```
master
- d102a4cb41d663bd43f482b1f69a1e4c9c98d361 - tag: v1.0.1 --> trigger korlibs-version@master klockVersion=1.0.1
kotlin-1.4
- 2fd4a1696e3fac07fde04464f6c65ed4a9b7ccd4 - tag: v2.0.3-alpha --> trigger korlibs-version@kotlin-1.4 klockVersion=2.0.3-alpha
```

emign: When triggering a workflow via tag, `GITHUB_REF` does not contain the actual branch name.
I haven't yet found a way to actually get the branch name

Reference:
https://github.community/t/get-the-branch-name/18225

soywiz: Check #korgeorg

PR #263, opened 2020-07-10
Disable DSL scope limiting by removing KorDslMarker from View.

### LINKED ISSUE #262: Issues with the DSL scope limiting
Cause issues on korge-samples at unexpected places where it could be annoying. We should try to fix that or disable the scope limiting.

soywiz:
Created an annotation that now is a dummy annotation but that can be changed to an typealias for KorDslMarker so we can switch to it fast and check stuff

Fixes #262


soywiz:
Asked for help here:
https://discuss.kotlinlang.org/t/dsl-scope-limiting-with-two-different-dsls/18404

soywiz:
Video:
https://www.youtube.com/watch?v=ljbnifyZS70

PR #273, opened 2020-07-14
Fix overlay proposal

### LINKED ISSUE #164: Add fps and graph overlay

soywiz:
Fixes #164

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/FpsOverlay.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/FpsOverlay.kt
@@ -57,10 +57,11 @@ internal fun ViewsContainer.installFpsDebugOverlay() {
         )
 
         for (n in 0 until frequencies.size) {
-            val y = (overlayHeight - frequencies[n].toDouble().convertRange(
+            val scaledFreq = frequencies[n].toDouble().convertRange(
                 minFps.toDouble() - 5.0, maxFps.toDouble(),
                 overlayHeightGap, overlayHeight
-            ) + graphTop).toFloat()
+            )
+            val y = (graphTop + overlayHeight - scaledFreq).toFloat()
             val x = graphLeft + (n.toFloat() * overlayWidth / MAX_FREQUENCIES_SIZE.toFloat())
             if (n > 0) {
                 renderContext.debugLineRenderContext.line(previousX, previousY, x, y)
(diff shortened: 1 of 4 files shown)


korge/src/commonMain/kotlin/com/soywiz/korge/view/FpsOverlay.kt
```diff
@@ -0,0 +1,77 @@
...
+        renderContext.debugLineRenderContext.line(
+            graphLeft, graphTop + overlayHeight.toFloat(),
+            graphLeft + overlayWidth.toFloat(), graphTop + overlayHeight.toFloat()
+        )
+
+        for (n in 0 until frequencies.size) {
+            val y = (overlayHeight - frequencies[n].toDouble().convertRange(
+                0.0, maxFps.toDouble(),
+                overlayHeightGap.toDouble(), overlayHeight.toDouble()
+            ).toFloat() + graphTop).toFloat()
```
soywiz, line 68: I think we should move graphTop to the start of the expression

soywiz: Like `graphTop + overlayHeight - ... expr` . And we should move expr to another variable I guess

RezMike: Just add this fix in another commit

korge/src/commonMain/kotlin/com/soywiz/korge/view/FpsOverlay.kt
```diff
@@ -10,26 +10,22 @@ import com.soywiz.korim.color.*
 internal fun ViewsContainer.installFpsDebugOverlay() {
     val frequencies = Deque<Int>()
```
soywiz, line 11: KDS should expose `IntDeque` that doesn't do boxing

korge/src/commonMain/kotlin/com/soywiz/korge/view/FpsOverlay.kt
```diff
@@ -0,0 +1,73 @@
...
+internal fun ViewsContainer.installFpsDebugOverlay() {
+    val frequencies = Deque<Int>()
+    var previousTime = DateTime.now()
+    var frames = 0
+    val MAX_FREQUENCIES_SIZE = 50
+    views.debugHandlers.add { ctx ->
+        val scale = ctx.ag.devicePixelRatio
+
+        val fontSize = 8.0 * scale
+        val currentTime = DateTime.now()
```
soywiz, line 19: We should use `PerformanceCounter` to have sub-millisecond precision

PR #274, opened 2020-07-19
Add timePerFrame to SpriteAnimation's companion invoke

MScottMcBee:
SpriteAnimation's companion invoke was the only way to create a SpriteAnimation without a way to set the frametime. This just adds it in.


RezMike:
Nice. Thanks!

PR #312, opened 2020-12-22
Add onCollisionExit extension functions to View

QuentinWidlocher:
Hello, this is my first open source PR so feel free to tell me if I'm doing something wrong here.

I use Unity a lot and I thought Korge lacked a collision exiting function. I think this is really useful to avoid using hitTest inside an updater to do it manually, and you cannot use shapes with this method.
Example case : detecting when a player is falling down a platform

I hope you'll consider it :blush:


QuentinWidlocher:
I don't understand why I have errors like "Unresolved reference: readVectorImage", considering my changes :thinking:

soywiz:
image

Perfectly valid. Thanks!

soywiz:
Those errors are my fault. Since I backported changes from korge-next but not bumped korim yet. Don't worry :)

Issue #307, opened 2020-12-16
[Question/request] 3D animation in function of progress, not time 


krzema12:
Hi,

I have a following use case. Imagine a steering wheel in a car: it rotates, and it can have its extreme left angle and extreme right angle. I want to be able to define in the model exported to a COLLADA format how a given movement looks, and then set its "progress" from the code. For example, if the progress is a value between 0 and 1, passing 0 would display the steering wheel in its leftmost position, 0.5 would mean a steering wheel of a car going straight, 0.75 would be slightly turned to the right.

I'm looking at `com.soywiz.korge3d.animation.Animator3D`, but it doesn't seem to support such use case. It just plays back the animation as the time flows. I'm wondering if animations as such are a proper tool for my use case. Currently I implement such stuff by rotating a given part of the model manually, but I see two drawbacks:

1. The code is too tightly coupled with the model. Some parameters set from the code may not be correct anymore when changing some positions of elements in the model.
2. Not trivial to define e.g. rotation axis, so rotating by a single axis is not always possible. It's possible to define compound rotation, but then we're sort of back to the problem 1.

Please advice, and maybe consider adding support for something like this to `Animator3D`? :) Even selecting the frame by its ordinal number would be helpful!

Best,
Piotr

soywiz:
3D is planned for 3.0. But if you have ideas on how to implement this properly, PRs are welcome before 3.0!

krzema12:
When are you planning to release 3.0? I'm asking to assess if I can wait or try to contribute :) Or at least implement some variation of the animator logic in my project, and then share it with you. I need this feature for my project.

soywiz:
No ETA. It hugely depends on the amount of time I'm able to spend on this myself and/or the amount of contributions we get on this. From the initial work to 1.0 it took almost two years, and from 1.0 and 2.0 another two years. That doesn't mean that 3.0 will be need another two years, but probably will take at least 1 year to complete. So if you want results earlier, I'd suggest you to consider contributing, and that will also be greatly appreciated too! :)

krzema12:
A small heads-up: I'm working on a feature I call "playback patterns", which will allow the behavior I need from korge, and add a lot of flexibility to how animations are played like: oscillation, ease in/out, and any custom patterns expressed as a math function of one variable. I'm still iterating on the API design. I'm expecting to send some first iteration for review by the end of the year :)

soywiz:
Cool :) I'm keen for it 👍

[...]

PR #313, opened 2020-12-23
[#307] Make Animator3D handle single Animation

krzema12:
Currently for 3D animations it's only possible to play a collection
of them in a loop, with regular speed. The goal of this change is to
prepare the API for more fine-grained control over a single animation:
customizing playback speed, reversing playback, setting arbitrary progress
and more. That's why I recommend changing Animator3D's API to handle
a single animation instead of a collection.

I'm planning to extend the API as described above in the subsequent commits,
and would like this breaking change to go as a separate PR for better
visibility. Further changes will be backward-compatible.

This is a breaking API change, but given that KorGE 3D is still
experimental, I presume it's fine to just mention it in the release
notes. See a change in `korge/src/jvmTest/kotlin/com/soywiz/korge/korge3d/main.kt`
for a migration example.


soywiz:
image

PR #320, opened 2021-01-03
Partial fix for UIComboBox

jfbilodeau:
Partial fix for UIComboBox where the itemsList would cover underlying controls even when hidden

https://github.com/korlibs/korgw/issues/40

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/ui/UIComboBox.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/ui/UIComboBox.kt
@@ -16,44 +16,45 @@ inline fun <T> Container.uiComboBox(
 ) = UIComboBox(width, height, selectedIndex, items, verticalScroll, skin).addTo(this).apply(block)
 
 open class UIComboBox<T>(
-	width: Double = 192.0,
-	height: Double = 32.0,
-	selectedIndex: Int = 0,
-	items: List<T> = listOf(),
-	verticalScroll: Boolean = true,
+    width: Double = 192.0,
+    height: Double = 32.0,
+    selectedIndex: Int = 0,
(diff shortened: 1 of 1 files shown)


korge/src/commonMain/kotlin/com/soywiz/korge/ui/UIComboBox.kt
```diff
@@ -62,62 +63,72 @@ open class UIComboBox<T>(
...
 
     fun open() {
-        showItems = true
+        addChild(itemsView)
     }
 
     fun close() {
-        showItems = false
+        removeChild(itemsView)
     }
```
soywiz, line 78: Maybe, instead of mutating `showItems` variable, could it be an immutable property like `val shownItems get() = itemsView.parent == this` ?

jfbilodeau: @soywiz: Sounds good to me!

jfbilodeau: @soywiz: Sounds good to me!

soywiz: Can you change it in this PR?

jfbilodeau: Hey @soywiz,

I've looked at the code and it turns out the showItems var is necessary. It's used to determine if the dropdown should show instead of if it's visible. We'll have to leave as-is.

BTW, the assignments to showItems in the above code are removed from the PR

J-F

soywiz:
4t1dlo

Thanks!

PR #343, opened 2021-03-03
Add FlagFilter

MikeSmithEU:
A FlagFilter that distorts the texture using increasing waves to the right, keeping the left-most vertical static, as if connected to a flag pole.

Example of this filter in use:

https://user-images.githubusercontent.com/2065591/109841345-3d6de080-7c49-11eb-9cff-9f47da1f3162.mov

I have read CONTRIBUTING.md and accept contributing this to the korge project under the Apache 2.0 license...


MikeSmithEU:
Needs to be in korge-next

PR #430, opened 2022-05-25
Update README to mention korge-next

cbreezier:
When newcomers find the korge repo with all of the stars and issues, it can be off-putting to think that there have been no changes or releases for almost a year. Updating the readme to mention active development in korge-next can help alleviate this issue.


soywiz:
Thank you!

Issue #504, opened 2021-11-07
Update ASE asesprite to support 1.3 : TileMap/TileSet


soywiz:
https://github.com/aseprite/aseprite/commit/95e6b5ddd6eed2ab80f8d2f19109ba4d9931ad5b

https://github.com/aseprite/aseprite/blob/b7d5d4a2c9ea7ed4c9f7bc9ba665039d808ad3b6/docs/ase-file-specs.md

PR #511, opened 2022-06-30
Aseprite tilemap support

soywiz:
TileMap/TileSet functionality moved here: https://github.com/korlibs/korge/commit/28f37ce42065757566299bd4057f28fa54819583

Implemented the spec diff from here: https://github.com/aseprite/aseprite/commit/95e6b5ddd6eed2ab80f8d2f19109ba4d9931ad5b

Fixes https://github.com/korlibs/korge/issues/504


soywiz:
Still misses support on KorGE's side

soywiz:
Screenshot 2022-06-30 at 23 05 44

soywiz:
@jobe-m can you have a look? Since the view includes something related to view recycling Im not sure if the proposed implementation would work for you

jobe-m:
Cool, yes, I will have a look!

korim/src/commonMain/kotlin/com/soywiz/korim/format/ASE.kt
```diff
@@ -59,24 +65,54 @@ object ASE : ImageFormatWithContainer("ase") {
...
+        val x: Int
+        val y: Int
+        val opacity: Int
+        fun resolve(): AseCell = this
+    }
+    open class AseBitmapCell(
+        override val bmp: Bitmap,
+        override val x: Int, override val y: Int, override val opacity: Int,
+    ) : AseEntity by AseEntity(), AseCell
+    open class AstLinkedCell(
```
soywiz, line 79: Typo: ast->ase

jobe-m:
I checked the parallax-scrolling-aseprite example as it uses also classes which you changed here. It look somewhat good as it still shows all layer objects but I guess there is a -1/+1 problem with repeating the TileMap (IIRC) objects. Every first object from the left is not shown and it starts to appear when the full object is visible on the screen while moving to the right side. See picture:

image

korge/src/commonMain/kotlin/com/soywiz/korge/view/animation/ImageAnimationView.kt
```diff
@@ -111,15 +131,26 @@ open class ImageAnimationView<T: SmoothedBmpSlice>(
...
         nframes = animation?.frames?.size ?: 1
         // Before clearing layers let parent possibly recycle layer objects (e.g. return to pool, etc.)
-        for (layer in layers) { onDestroyLayer?.invoke(layer) }
+        for (layer in layers) {
+            if (layer is TileMap) {
+                onDestroyTilemapLayer?.invoke(layer)
+            } else {
+                onDestroyLayer?.invoke(layer as T)
+            }
+        }
```
jobe-m, line 140: LGTM!

korge-sandbox/src/commonMain/kotlin/Main.kt
```diff
@@ -94,6 +94,7 @@ suspend fun main() = Korge(
             Demo(::MainTilemapWithScroll),
             Demo(::MainTiledBackground),
             Demo(::MainAseprite),
+            Demo(::MainAseParallaxSample),
```
jobe-m, line 97: 👍

soywiz:
So the parallax sample seems to be broken, but doesn't seems to be related to this right? I believe it was broken when optimizations regarding visibily were done (for the other sample, the one that displays a huge tilemap with a donut), I have tried manually adjusting min/maxXY in the tilemap rendering and seems to paint stuff.
I have open a separate issue to investigate that: https://github.com/korlibs/korge/issues/512
Maybe we should also consider refactoring it, because it is getting a bit complicated

soywiz:
BTW the `SingleTile` view is only used to repeat an image in the parallax code, or are you using it somewhere else? If I refactor tilemaps and parallax to not require that class, would be fine for you?

PR #842, opened 2022-07-16
Fix zip file can't be read when uncompressed size is not in header

### LINKED ISSUE #841: Zip file can't be read when uncompressed size is not in header
I came across a valid .zip file that `ZipVfs` fails to read. I found it's because the uncompressed size is 0 in the header, which is used as the length when creating the input stream. The entry uncompressed size value should be used instead.

jeffdgr8:
I came across a valid .zip file that `ZipVfs` fails to read. I found it's because the uncompressed size is 0 in the header, which is used as the length when creating the input stream. The entry uncompressed size value should be used instead.

Fixes #841


soywiz:
Awesome. Thanks for reporting the bug, fixing it and adding the test :)

PR #860, opened 2022-07-22
Update README.md

andsiu:
Simple change from "github" to "GitHub"

No reference issue because this problem is too small to justify making one.


soywiz-invideo:
Thanks!

PR #879, opened 2022-07-26
require that forces applied to body are valid

jbellis:


kbox2d/src/commonMain/kotlin/org/jbox2d/collision/AABB.kt
```diff
@@ -308,9 +308,7 @@ class AABB {
                 return false
             }
 
-            return if (a.lowerBound.x - b.upperBound.x > 0.0f || a.lowerBound.y - b.upperBound.y > 0.0f) {
-                false
-            } else true
+            return !(a.lowerBound.x - b.upperBound.x > 0.0f || a.lowerBound.y - b.upperBound.y > 0.0f)
```
soywiz, line 311: We could simplify this furhter to avoid the negation.

kbox2d/src/commonMain/kotlin/org/jbox2d/collision/broadphase/DynamicTree.kt
```diff
@@ -74,7 +74,7 @@ class DynamicTree : BroadPhaseStrategy {
                     continue
                 }
 
-                assert(node.child1 == null == false)
+                assert(!(node.child1 == null))
```
soywiz, line 77: Ditto: We could simplify this furhter to avoid the negation.

soywiz (approved):
LGTM. We should add tests for kbox2d at some point to ensure it continues working at runtime after refactors even if they are safe. Let's do this at some point: https://github.com/korlibs/korge/issues/891

PR #890, opened 2022-07-27
Try run Android tests on linux without hardware acceleration

soywiz:


soywiz-invideo:
It takes 15minutes more to complete, but also since the computer used is a Linux one, it starts earlier, so I'm going to merge this

PR #902, opened 2022-07-30
Bump Kotlin to 1.7.20-Beta and serialization to 1.4.0-RC

soywiz:


soywiz:
The test that used to fail only in windows from time to time with the new memory manager, failed here too (without the new memory manager in theory):
https://github.com/korlibs/korge/runs/7592515872?check_suite_focus=true#step:5:3769

PR #927, opened 2022-08-07
Korge Screenshot Testing Framework (JVM only)

Kietyo:
Preview: https://youtu.be/Ni4U-O4TT50


soywiz:
Cool! Nice idea, thanks for implementing it!
What prevents this from being multiplatform?
Would be nice if we could use this for ensuring rendering consistency. Maybe we can accept some kind of thresold with images to accept minor inconsistencies due to opengl shader accuracy
* Can we also detect if there is a `CI=true` environment variable to be fully headless in the CI and fail there? In the CI in some cases we are not headless like here: https://github.com/korlibs/korge/blob/c138ea25343dc864821ac749a2caf80358498a39/.github/workflows/TEST.yml#L157

Kietyo:
1. Not familiar enough with multiplatform to support it. Was hoping for you to port it after this gets submitted.
2. Sure that could be done in a future PR.
3. Not that familiar with gradle environment variables.

My goal for this PR is to setup the basic framework and hope that you could support (1) and (3).

korge/src/jvmMain/kotlin/com/soywiz/korge/testing/testing_utils.kt
```diff
@@ -0,0 +1,339 @@
...
+) {
+    System.setProperty("java.awt.headless", "false")
+    val throwable = Throwable()
+    val testClassName = throwable.stackTrace[0].className
+    val testMethodName = throwable.stackTrace[0].methodName
+    val existingMain = korgeConfig.main
+    val results = KorgeTestResults(testMethodName)
+    // Keep locked while tester is running
+    val testingLock = Mutex(locked = true)
+    korgeConfig.main = {
```
soywiz, line 217: To avoid making the configuration mutable. You can do something like:

```kotlin
val finalKorgeConfig = korgeConfig.copy(main = {
 ...

Then:

```kotlin
Korge(finalKorgeConfig)
```
})
```

File: korge/src/jvmMain/kotlin/com/soywiz/korge/testing/testing_utils.kt [hunk 2]
```diff
@@ -0,0 +1,339 @@
...
+
+    companion object {
+        val DATE_FORMAT = DateFormat("yyyy-dd-MM HH:mm:ss z")
+        val PATH_DATE_FORMAT = DateFormat("yyyyMMdd_HH_mm_ss_z")
+        private val LINE_BREAK_WIDTH = 100
+    }
+}
+
+@OptIn(KorgeExperimental::class)
+inline fun Any.korgeTest(
```
soywiz, line 205: Why `fun Any.korgeTest()` and not just `fun korgeTest()`

korge/src/jvmMain/kotlin/com/soywiz/korge/testing/testing_utils.kt
```diff
@@ -0,0 +1,339 @@
...
+        }
+    }
+}
+
+data class KorgeTestResults(
+    val testMethodName: String,
+    val results: MutableList<KorgeTestResult> = mutableListOf()
+)
+
+class KorgeTester(
```
soywiz, line 84: Can we rename it to `KorgeScreenshotTester` as well as the `korgeTest` -> `korgeScreenshotTest`?

korge/src/commonMain/kotlin/com/soywiz/korge/Korge.kt
```diff
@@ -587,7 +587,7 @@ object Korge {
         val quality: GameWindow.Quality? = null,
         val icon: String? = null,
         val multithreaded: Boolean? = null,
-        val main: (suspend Stage.() -> Unit)? = module.main,
+        var main: (suspend Stage.() -> Unit)? = module.main,
```
soywiz, line 590: As described below, this can be kept as immutable by using `.copy()`

korge/src/jvmMain/kotlin/com/soywiz/korge/testing/testing_utils.kt
```diff
@@ -0,0 +1,339 @@
...
+        existingMain?.invoke(this)
+        callback(this, korgeTester)
+    }
+    suspendTest {
+        try {
+            Korge(korgeConfig)
+        } catch (exception: HeadlessException) {
+            // Running in a headless environment (e.g github tests).
+            // Return to mark as passing.
+            return@suspendTest
```
soywiz, line 230: It shouldn't be passing in CI. If it is in CI, we should compare screenshots and if they don't match, fail. Otherwise if we forget to run tests locally we would miss screenshot changes

Kietyo: we cant run it in the first place in CI since it's a headless env

korge/src/jvmMain/kotlin/com/soywiz/korge/testing/testing_utils.kt
Kietyo: Done

korge/src/jvmMain/kotlin/com/soywiz/korge/testing/testing_utils.kt
Kietyo: done

korge/src/jvmMain/kotlin/com/soywiz/korge/testing/testing_utils.kt
Kietyo: done

korge/src/commonMain/kotlin/com/soywiz/korge/Korge.kt
Kietyo: done

Kietyo:
Have a few more changes to make. Will let you know when ready to review again.

Kietyo:
Done. Ready now.

soywiz:
Thanks! Merged already. I will investigate later how we can make it work on the CI and in other targets. I have created this issue for that: https://github.com/korlibs/korge/issues/936

PR #928, opened 2022-08-08
Add the ability to output to pre-allocated bitmap for QOI and added benchmarks.

Kietyo:


soywiz:
I would suggest moving the extra parameter to the `ImageDecodingProps`: https://github.com/korlibs/korge/blob/c138ea25343dc864821ac749a2caf80358498a39/korim/src/commonMain/kotlin/com/soywiz/korim/format/ImageFormat.kt#L111

Something like optional nullable output bitmap that some codecs can reuse for single image decoding in the case dimensions & bpp match. For KorGE 3.0 I tried to remove as many extra parameters as possible for image/audio decoding, so that would be aligned with that. What do you think?

Kietyo:
Will do. Still WIP will let you know when it's ready.

Kietyo:
Messed up the git history for this PR, closing and making a new one.

PR #945, opened 2022-08-26
Add the ability to output to pre-allocated bitmap for QOI and added benchmarks. (Attempt 2)

Kietyo:


korim/src/commonMain/kotlin/com/soywiz/korim/format/TGA.kt
```diff
@@ -1,5 +1,6 @@
 package com.soywiz.korim.format
 
+import com.soywiz.korim.bitmap.Bitmap
```
soywiz, line 3: There are now files that have new imports but no other code change. Can you remove those imports so the files are not updated?

Kietyo: done

korim/src/commonMain/kotlin/com/soywiz/korim/format/ImageFrameLayer.kt
```diff
@@ -42,6 +43,7 @@ open class ImageFrameLayer constructor(
     val height get() = slice.height
     val area: Int get() = slice.area
     val bitmap: Bitmap get() {
+        if (returnBitmapInPlace) return slice.bmp
         if (_bitmap == null) _bitmap = slice.extract()
```
soywiz, line 47: Maybe instead of propagating the `returnBitmapInPlace` we can optimize this.
What if we check if the bounds of the slice are the whole bitmap, and if that's the case, we return the slice.bmp without extracting? That way it will work, be faster and will simplify by not having to propagate or have the returnBitmapInPlace property

Kietyo: done

korim/src/commonMain/kotlin/com/soywiz/korim/format/ImageFrameLayer.kt
```diff
@@ -18,14 +18,13 @@ import com.soywiz.korim.tiles.TileMapData
...
  */
-open class ImageFrameLayer constructor(
+open class ImageFrameLayer(
     val layer: ImageLayer,
     slice: BmpSlice,
     val targetX: Int = 0,
     val targetY: Int = 0,
     val main: Boolean = true,
     val includeInAtlas: Boolean = true,
-    val linkedFrameLayer: ImageFrameLayer? = null,
```
soywiz, line 28: Why is this removed?

korge/src/commonTest/kotlin/com/soywiz/korge/view/ImageAnimationViewTest.kt
```diff
@@ -13,23 +13,63 @@ import kotlin.test.assertEquals
 class ImageAnimationViewTest {
     private val animImages4Layers1 = ImageAnimation(
         frames = listOf(
```
soywiz, line 15: Is this file reformatted only? Or changes something? Because since it has a lot of things I cannot say for sure

korim/src/commonMain/kotlin/com/soywiz/korim/format/ASE.kt
```diff
@@ -564,7 +572,15 @@ object ASE : ImageFormatWithContainer("ase") {
                     maskRotate = resolved.bitmask90CWFlip,
                 )
             }
-            return ImageFrameLayer(layer, resolved.bmp.slice(), resolved.x, resolved.y, main = false, includeInAtlas = true, tilemap = tilemap)
+            return ImageFrameLayer(
```
soywiz, line 575: This looks like a reformat. Should it be included in this PR?

korim/src/commonMain/kotlin/com/soywiz/korim/format/ASE.kt
```diff
@@ -619,7 +635,7 @@ object ASE : ImageFormatWithContainer("ase") {
                         it.layer, it.slice.slice(sliceFrame, it.slice.name),
                         (if (props.useSlicePosition()) sliceKey.sliceFrame.x else it.targetX) - sliceKey.pivotX,
                         (if (props.useSlicePosition()) sliceKey.sliceFrame.y else it.targetY) - sliceKey.pivotY,
-                        it.main, it.includeInAtlas, null
+                        it.main, it.includeInAtlas
```
soywiz, line 638: This null parameter was the property that was removed? It was not used and always null in the end?

korim/src/commonMain/kotlin/com/soywiz/korim/format/ImageFormat.kt
```diff
@@ -74,6 +74,8 @@ abstract class ImageFormat(vararg exts: String) : ImageFormatDecoder {
 
 	fun decode(s: SyncStream, props: ImageDecodingProps = ImageDecodingProps.DEFAULT): Bitmap = this.read(s, props)
 	//fun decode(file: File, props: ImageDecodingProps = ImageDecodingProps()) = this.read(file.openSync("r"), props.copy(filename = file.name))
+
+    // Decodes a given byte array to a bitmap based on the image format.
```
soywiz, line 78: To have documentation generation and shown in the IDE

```suggestion
    /** Decodes a given byte array to a bitmap based on the image format. */
```

korim/src/commonMain/kotlin/com/soywiz/korim/format/ImageFormat.kt
```diff
@@ -120,6 +122,10 @@ data class ImageDecodingProps constructor(
     val preferKotlinDecoder: Boolean = false,
     val tryNativeDecode: Boolean = true,
     val format: ImageFormat? = RegisteredImageFormats,
+    // Provides an `out` parameter to reuse an existing Bitmap to reduce allocations.
+    // Note though that not all formats may use the bitmap provided by the `out` param.
+    // In those cases, they will return a newly allocated Bitmap instead.
+    val out: Bitmap? = null,
```
soywiz, line 128:
```suggestion
    /**
     * Provides an `out` parameter to reuse an existing Bitmap to reduce allocations.
     *
     * Note though that not all formats may use the bitmap provided by the `out` param.
     * In those cases, they will return a newly allocated Bitmap instead.
     */
    val out: Bitmap? = null,
```

korim/src/commonMain/kotlin/com/soywiz/korim/format/ImageFrameLayer.kt
Kietyo: unused

korim/src/commonMain/kotlin/com/soywiz/korim/format/ImageFrame.kt
```diff
@@ -33,32 +33,51 @@ open class ImageFrame(
...
             index: Int = 0,
-        ): ImageFrame = ImageFrame(index, time, fastArrayListOf(ImageFrameLayer(
-            ImageLayer(0, null, ImageLayer.Type.NORMAL),
-            bitmap.slice(name = name),
-            targetX,
-            targetY,
-            main,
-            includeInAtlas,
-        )))
+        ): ImageFrame =
```
soywiz, line 36: Only reformat file?

korim/src/commonMain/kotlin/com/soywiz/korim/format/KRA.kt
```diff
@@ -223,15 +223,21 @@ object KRA : ImageFormat("kra") {
                                         layerPartialBitmap,
                                         targetX = layerX,
                                         targetY = layerY,
-                                        name = layerName,
-                                        main = false
+                                        main = false,
+                                        name = layerName
```
soywiz, line 227: Was this a cleanuo or something from the IDE to keep params sorted or something?

korim/src/commonMain/kotlin/com/soywiz/korim/format/ImageFrame.kt
Kietyo: yea

korim/src/commonMain/kotlin/com/soywiz/korim/format/QOI.kt
```diff
@@ -31,13 +30,26 @@ object QOI : ImageFormat("qoi") {
         }
     }
 
+    // QOI supports outputting to the `out` bitmap, but certain conditions must be met.
+    // 1. The width and height of the output Bitmap must match the header width and height.
+    // 2. The out Bitmap must be a Bitmap32.
```
soywiz, line 35:
```suggestion
    /*
     * QOI supports outputting to the `out` bitmap, but certain conditions must be met.
     * 1. The width and height of the output Bitmap must match the header width and height.
     * 2. The out Bitmap must be a Bitmap32.
     */
```

korge/src/commonTest/kotlin/com/soywiz/korge/view/ImageAnimationViewTest.kt
Kietyo: auto reformatted

korim/src/commonMain/kotlin/com/soywiz/korim/format/ASE.kt
Kietyo: I just auto format with the IDE bot worth creating separate PR imo

korim/src/commonMain/kotlin/com/soywiz/korim/format/ASE.kt
Kietyo: yea

korim/src/commonMain/kotlin/com/soywiz/korim/format/KRA.kt
Kietyo: looks like it was auto reformatted from ide

soywiz:
LGTM.

Can you change the `//` comments for methods to `/** */` doc comments, so they act as documentation? Using `[param]` to refer to parameters.
Are the reformats intended? And the parameter/property removal?

After using KorGE a bit with all the explicit imports I found it pretty inconvenient to use due to extension methods and stuff, so I changed it in the template: https://github.com/korlibs/korge-hello-world/blob/87aed7e99113ebfaac3da6fc9c3a67a345728493/.editorconfig#L7
When having a lot of properties in a class I find putting them in a single line convenient because it is easier for me to read them, while when formatted it splits in 4 or 5 lines and then I lose a lot of context and it is harder for me to find the differences and what's the same. Do you know if there is a way to keep those properties as they are while being able to reformat?

Kietyo:
Reformats are intended, I usually instinctively format after modifying things.

> After using KorGE a bit with all the explicit imports I found it pretty inconvenient to use due to extension methods and stuff,

I don't really understand why it would be inconvenient? If you have method A that's implemented in both package A and package B, you'll know right away which package is being used. With wildcard, you have no idea.

> When having a lot of properties in a class I find putting them in a single line convenient because it is easier for me to read them, while when formatted it splits in 4 or 5 lines and then I lose a lot of context and it is harder for me to find the differences and what's the same. Do you know if there is a way to keep those properties as they are while being able to reformat?

I'm down with whatever. Something you could do is share your intellij kotlin code style settings:

image


I am currently using Google's style:
https://github.com/google/styleguide/blob/gh-pages/intellij-java-google-style.xml

Which can be imported here:
image

soywiz:
For example, when using tweens, if using the star import, you get all the get extension methods while otherwise you have to manually import everything, and since they are extension getters it is less convenient.

When you have a extension member in several files, intellij allows you to select the one to be used for this file, and you can even use aliases.
I believe JetBrains used to prefer the all `import *` approach, while google used to prefer the single import approach. I prefer the wildcard import approach, specially after trying it a bit.

I will try to figure out if there are some configurations that wouldn't touch the single-line properties. In general I prefer editing the `.editorconfig` and the `ij_` prefixes since it is autoconfigured when opening the project without having to import anything.

ie.: https://github.com/korlibs/korge/blob/4fefa10bc4e75fed52567c2dd1f5063cbdb6d9d7/.editorconfig#L7

soywiz:
Thanks!

PR #966, opened 2022-09-05
Optimization: reduce overhead when calling components, support rendering only when something changed, & WIP code for optimizing event listening too

soywiz-invideo:
* Now instead of iterating all the views for the components, it goes to the nodes with components of a specified kind
* Now instead of rendering always, there is a new mode where we can only render if something changed
* WIP: New code for event listeners based on the same concept for components


soywiz:
@TODO: Add tests verifying that changing all the properties trigger a re-render in next frame

soywiz-invideo:
We might also consider copying the views or the stage instance to each view individually when attached to the stage, so we don't have to iterate over the ancestors for each property change.

Issue #959, opened 2022-08-30
[klock] clarify that DateTime.fromUnix takes milliseconds instead of seconds


MinmoTech:
Currently `DateTime.fromUnix()` takes a timestamp in milliseconds but with just the name `fromUnix` I would have expected it to use seconds instead.

I would recommend to rename this function to `fromUnixMilliseconds` and maybe to additionally add the function `fromUnixSeconds`.

soywiz:
Yeah, makes sense: in PHP, Python and C I believe unix are in seconds. While in Java and JS it is in milliseconds.
Can you make a PR? Maybe the method can be deprecated and suggest a replacement to use `fromUnixMilliseconds` instead

Kietyo:
fixed

PR #975, opened 2022-09-07
Create new methods which detail the unix unit for time apis and add deprecation annotations to prefer the new methods.

Kietyo:
-


soywiz:
Fixes https://github.com/korlibs/korge/issues/959

soywiz:
Yes, sorry. Forgot about it.

Thanks!

PR #988, opened 2022-09-16
Update korge-fleks to version 1.6-KMP

jobe-m:
The sources in korge-fleks module were updated to Fleks upstream version 1.6-KMP. The sample sources in korge-sandbox were updated to use the new DSL notation for creating a world.

Change log for Fleks since last update:
- BUGFIX: fix an issue that entities did not get removed from families that only used the noneOf configuration. The reason is that a removed entity has no components which is then actually valid for such families. However, a removed entity should of course be removed from any family ;)
- UPDATE: BitArray has a new toString function which makes it easier to debug certain things.
- NEW: new snapshot and snapshotOf function for the World to get a snapshot of the entire world (=entities and components) or of a specific entity. This should also help with debugging. A new page to the Fleks Wiki was added as well regarding debugging.
- NEW: new loadSnapshot function for the World to load a specific snapshot. This will clear the current world and create entities and components according to the provided snapshot.
- BUGFIX: fixed numEntities of a family to return the correct number of entities instead of the highest ID of any entity inside the family
- NEW: added isEmpty, isNotEmpty, first and firstOrNull functions to a family
- BUGFIX: Entities created inside a system's constructor are now properly added to their family and also EntityListener get notified correctly in this case
- BUGFIX: ComponentListener did not get notified when entities are created in a system's init block
- UPDATE: the used parameter of an injectable is removed because injectables are now properly resolved and cleared once the world is created
- UPDATE: minor memory management improvements
- NEW: make world a default dependency for the dependency injection mechanism. That way it is now possible to easily pass the world to a ComponentListener
- NEW: add getter for world's systems
- NEW: add configureEntity to world. This makes it now possible to modify an entity outside of a system as requested in the comments section of my YouTube channel
- NEW: add support to create families out of a system via the world. This is the biggest change of this version and adds new flexibility on how to handle and iterate over entities with specific component configurations
- NEW: users have now access to the family of an IteratingSystem
- NEW: family now also contains a configureEntity function which allows users to modify an entity outside of a system in a convenient way
- NEW: a new FamilyListener support is added which allows users to react on when an entity is added to, or removed from a family
- NEW: new DSL to create a world. Use components , systems, injectables or families block to define your world. For more details, refer to the ReadMe
- NEW: the KMP setup was updated to a proper multiplatform gradle setup. That way it is now possible to add Fleks to a KorGE dependency via following line in the korge block:

    dependencyMulti("io.github.quillraven.fleks:Fleks:1.6-KMP",
    registerPlugin = false)


jobe-m:
@soywiz  I wanted you to acknowledge that the korge-fleks module is updated with latest content from the Fleks repository.
I guess we concluded that we want to keep a copy of Fleks ECS in Korge in case when we need a platform build of it which the Fleks upsteam project does not provide.
Is it ok to merge this PR?

BTW how can I trigger CI to build the unit tests on GitHub? I ran them locally and they passed but wanted to be sure that it does not break on CI :)

soywiz:
Awesome! Nice work 👍

PR #1018, opened 2022-10-04
Increase AsyncInjector coverage

soywiz:


soywiz:
I think there is a strange issue here of some files having different result (partial lines) in different runs. Not sure if a misconfiguration somewhere on KorGE's side or something else.

PR #1067, opened 2022-10-24
Fix bug in BmpSlice.extractWithBase()

Stechmuck:
`BmpSlice.extractWithBase()` returns Bitmaps with wrong content, if BmpSlice is rotated


soywiz:
Thanks!

PR #1069, opened 2022-10-26
Moved all korge modules into separate repositories with kprojects

soywiz:
Also updated https://github.com/korlibs/korge-hello-world to use kproject


jobe-m:
I am just trying it out and check how kprojects is working :D

soywiz:
@jobe-m So, while working on updating it you would open the example project:

https://github.com/korlibs/korge-fleks/tree/main/example

Then you would toggle this line and comment the git one:
https://github.com/korlibs/korge-fleks/blob/752c0933e36bcf785d363c07a1fba386f7c05203/example/deps.kproject.yml#L5

Then it should work as a normal local module. Then you can commit and use the new commit right away without compiling/uploading to maven

jobe-m:
When uncommenting line 5 (../korge-fleks) and commenting the "git" line I can successfully build under windows.
When trying to use the "git" line I get the following error:

```
Settings file 'D:\home\korge-fleks\example\settings.gradle.kts' line: 3

An exception occurred applying plugin request [id: 'com.soywiz.kproject.settings', version: '0.0.2']
> Failed to apply plugin 'com.soywiz.kproject.settings'.
   > Invalid repo 'korlibs/korge-fleks'
```
I quickly looked up in KProjectModel.kt the following lines:

```
fun normalizePath(path: String): String = File(path).normalize().toString()
fun ensureRepo(repo: String): String = normalizePath(repo).also { if (it.count { it == '/' } != 1) error("Invalid repo '$repo'") }
``` 
Could it be that on windows the `normalize()` function from java.io converts `/` to `\`? I haven't checked that myself...

jobe-m:
I just checked my assuption in kproject repo and it seems to be right ;)
I printed out the `it` value in `ensureRepo` and it looks like that (in brackets)

```
   > Invalid repo 'korlibs/korge-fleks' (korlibs\korge-fleks)
```
Let me quickly fix that and create a PR :)

jobe-m:
@soywiz FYI

soywiz:
Ohh, thanks for figuring out. I only tried on mac. Since I was using JGit I thought it would work on windows too, but I guess the \ and / mismatches is going to be always an issue  ¯\_(ツ)_/¯

I will probably add e2e tests when fixed to run on windows too

jobe-m:
@soywiz I cannot push directly to korlibs/kproject.git git repo for some reason. So I created a fork for pushing the fix in a PR.

PR #1095, opened 2022-11-07
Prototype new animator, supporting attaching new nodes on the fly

soywiz:


jobe-m:
Is there an example to check/learn how this works? :)

soywiz:
@jobe-m https://github.com/korlibs/korge-mascots/tree/main/example

PR #1098, opened 2022-11-09
Allow using a pre-allocated array for QOI encoding to avoid allocating a new array for every call.

Kietyo:


korim/src/commonTest/kotlin/com/soywiz/korim/format/QOITest.kt
```diff
@@ -120,4 +163,18 @@ class QOITest {
...
+    @Test
+    fun providingSmallPreAllocatedArrayResultsInError() = suspendTestNoBrowser {
+        val original = QOI.decode(resourcesVfs["dice.qoi"])
+        val preallocatedArray = UByteArrayInt(4)
+        val props = ImageEncodingProps()
+        props.preAllocatedArrayForQOI = preallocatedArray
+        assertFailsWith<IllegalArgumentException> {
+            QOI.encode(
+                original,
+                props
```
soywiz, line 176: To keep it clean and one-liner when using external properties, I typically use apply (so you dont need to create a local variable and choose a name):

```
ImageEncodingProps().apply {
         preAllocatedArrayForQOI = preallocatedArray
}
```

soywiz: Sometimes I add to those class an optional extra lambda or a new constructor with a lambda that automatically calls it, so it is even cleaner:

```
ImageEncodingProps {
         preAllocatedArrayForQOI = preallocatedArray
}
```

```
class ImageEncodingProps(…, init: (ImageEncodingProps.() -> Unit)? = null) {
init { init?.invoke() }
}

File: korge-benchmarks/src/jvmMain/kotlin/com.soywiz.benchmarks/QOIBenchmarks.kt [hunk 2]
```diff
@@ -45,6 +52,22 @@ class QOIBenchmarks {
...
         //  CI (99.9%): [1015.761, 1033.624] (assumes normal distribution)
-        return QOI.decode(qoiBytes, prop)
+        return QOI.decode(qoiBytes, decodePropToOutputBitmap)
+    }
+
+    @Benchmark
+    fun encodeBitmap(): ByteArray {
+        //2257.462 ±(99.9%) 49.785 us/op [Average]
+        //(min, avg, max) = (2109.371, 2257.462, 2735.280), stdev = 146.792
+       ubled //CI (99.9%): [2207.677, 2307.247] (assumes normal distribution)
```
soywiz, line 62: I guess this `ubled` was not meant to be included

Kietyo:
rdy

korim/src/commonMain/kotlin/com/soywiz/korim/format/QOI.kt
```diff
@@ -117,98 +129,121 @@ object QOI : ImageFormat("qoi") {
...
         val index = RgbaArray(64)
-        val maxSize = QOI_HEADER_SIZE + (bitmap.width * bitmap.height * (4 + 1)) + QOI_PADDING_SIZE
-        val bytes = UByteArrayInt(maxSize)
+        val maxSize = calculateMaxSize(bitmap)
+        val bytes = if (props.preAllocatedArrayForQOI == null) {
+            UByteArrayInt(maxSize)
+        } else {
+            require(props.preAllocatedArrayForQOI!!.size >= maxSize) {
+                """
+                   Requires a pre-allocated array of at least $maxSize bytes.
```
soywiz, line 138: We might want to remove this string at some point. I have been thinking and there are a lot of error strings in the code that cannot be stripped out from the executable, so even if only a few bytes they are summing up. It would be nice if we had a mechanism to include some strings on debug executables and trim them from release ones like a static if or something (wishlist for Kotlin)

soywiz:
Lgtm. Thanks!

korim/src/commonMain/kotlin/com/soywiz/korim/format/QOI.kt
Kietyo: if trimming it means no more useful error messages than as a client i would rather have more useful error messages than a smaller binary size IMO

soywiz: The exception type or code should be enough. While debugging it might be useful the error, but in the final binary in korge at least, where that exception shouldn't happen it is only a string hanging on there, and if it happens, an error code should suffice for the original developer to find out the problem.
Maybe a kotlin flag with a static if to decide whether to include them if you want to make that tradeoff.
One of the cool things of KorGE is that it produces small, and self-contained executables (no external DLLs/frameworks/.so other than the operating system ones, not hundreds of megabytes)
Lot of design decisions have been made to allow the DCE/tree shaking to be able to do its work, and not including zillion of external libraries is also a design decision. Limiting included strings might be something desiree at some point, but when we can do that without compromising the debug experience. But in general you will see OutOfBounds exceptions with only the index that failed or as much as the size of the collection, qnd thats enoigh. So maybe here an OutOfBounds or NotEnoughCapacityException might also be enough without the full string

Kietyo: thx for the insight!

PR #1101, opened 2022-11-10
Added CachedContainer functionality + improve _stage lookup by propagating it

### LINKED ISSUE #974: Copy the Views/Stage singleton to each View instance when attached to the stage to optimize rendering invalidation

### LINKED ISSUE #1064: Support caching tree views into textures & make dirty updating to be aware of this

soywiz:
This functionality supports caching a container to store the children in a texture and render that texture instead of all the children. It is integrated with the render invalidation system, so if a descendant changes calls invalidateRender, the texture is invalidated and the children re-rendered.
This support rendering huge hierarchies that don't change often, or small hierarchies with costly filters, super quickly.

This PR also improves how stage and render invalidator parent is computed, so it won't require iterating ancestors to get it which should be quicker for already constructed hierarchies.

The demo shows a node that has 100,000 solid rect children. When rendering those 100,000 children everyframe, it is pretty slow, but when cached into a texture, it is super fast (only rendering one texture). This demo also updates 2000 children every second, so when that happens, the children will be re-rendered into that texture once per second, while being lightning fast the rest of the frames.

For example, when doing UI, we could cache every UIWindow, so re-ndering is amortized.

https://user-images.githubusercontent.com/570848/201120451-4be95465-286d-49b5-bd31-8ffba22cee7f.mp4

Fixes https://github.com/korlibs/korge/issues/1064
Fixes https://github.com/korlibs/korge/issues/974

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/CachedContainer.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/CachedContainer.kt
@@ -1,10 +1,7 @@
 package com.soywiz.korge.view
 
-import com.soywiz.kmem.*
 import com.soywiz.korge.baseview.*
 import com.soywiz.korge.render.*
-import com.soywiz.korge.view.filter.*
-import com.soywiz.korim.bitmap.*
 import com.soywiz.korio.lang.*
 import com.soywiz.korma.geom.*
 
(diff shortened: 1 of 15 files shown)


korge/src/commonMain/kotlin/com/soywiz/korge/view/CachedContainer.kt
```diff
@@ -0,0 +1,105 @@
...
+            val texHeight = (lbounds.height).toInt().coerceAtLeast(1)
+            //println("texWidth=$texWidth, texHeight=$texHeight")
+            cache.resize(texWidth, texHeight)
+            ctx.renderToFrameBuffer(cache.rb) {
+                @Suppress("DEPRECATION")
+                ctx.batch.setViewMatrixTemp(tempMat2d.also {
+                    it.copyFrom(globalMatrixInv)
+                    it.translate(-lbounds.x, -lbounds.y)
+                }) {
+                    renderChildrenInternal(ctx)
```
mayankmkh, line 70: Can be `super.renderInternal(ctx)`?

korge/src/commonMain/kotlin/com/soywiz/korge/view/CachedContainer.kt
```diff
@@ -0,0 +1,105 @@
...
+        if (dirty) {
+            lbounds.copyFrom(getLocalBoundsOptimizedAnchored(includeFilters = false))
+            dirty = false
+            val texWidth = (lbounds.width).toInt().coerceAtLeast(1)
+            val texHeight = (lbounds.height).toInt().coerceAtLeast(1)
+            //println("texWidth=$texWidth, texHeight=$texHeight")
+            cache.resize(texWidth, texHeight)
+            ctx.renderToFrameBuffer(cache.rb) {
+                @Suppress("DEPRECATION")
+                ctx.batch.setViewMatrixTemp(tempMat2d.also {
```
mayankmkh, line 66: Can replace deprecated `ctx.batch.setViewMatrixTemp` with `ctx.setViewMatrixTemp`

korge/src/commonMain/kotlin/com/soywiz/korge/view/CachedContainer.kt
```diff
@@ -0,0 +1,105 @@
...
+                    it.translate(-lbounds.x, -lbounds.y)
+                }) {
+                    renderChildrenInternal(ctx)
+                }
+            }
+        }
+
+        //println("cache.tex=${cache.tex}")
+
+        ctx.flush()
```
mayankmkh, line 77: useBatcher flushes automatically if the batcher changes. Is this flush required here?

korge/src/commonMain/kotlin/com/soywiz/korge/view/CachedContainer.kt
soywiz: Done

korge/src/commonMain/kotlin/com/soywiz/korge/view/CachedContainer.kt
soywiz: Done

Issue #1116, opened 2022-11-17
Optimize draw batches by using VAOs, UBOs, buffering, single buffer uploads and separate thread for rendering


soywiz:
While the new SDF-based UI looks really beautiful, introduces new performance requirements that weren't here before. The previous UI was using a 9-patch, and it was less flexible, but worked great along bitmap fonts, since the current KorGE batcher supports up to 4 textures in a single batch. So we could render buttons and text in a single batch call.
There were some introduced optimizations to cache whole sub-graphics, so we can for example cache a whole UI window, or the whole UI or part of it, so it doesn't affect the rest of the game updating, but still, we want to be faster.

Now, rendering the background of the button and the text itself are two separate batches. We need to set the vertices, the uniforms and the data for the background, and then go back to the text rendering.

Each batch is slow because requires setting the vertex data, attributes and uniforms. Doing a VisualVM profiling we can identify the hot points here:
![Screenshot 2022-11-16 at 14 43 16](https://user-images.githubusercontent.com/570848/202385524-c6998ad5-ab79-4fc5-a0b0-513c71cbb368.png)

Since we are already using lists that would allow to execute stuff in parallel, we can start optimizing some stuff while being future-proof for other backends.
The idea here is to use proper VAOs and UBOs and also buffer everything to the end of the list.
This will require either extensions available, or WebGL >= 2 and Open GL ES >= 3.0.
It is possible to read more about this here: https://webgl2fundamentals.org/webgl/lessons/webgl2-whats-new.html
* On desktop, we should have that functionality already.
* On iOS, that's already there for ages.
* On Android 6,7% is still having only 2.0 available: https://developer.android.com/about/dashboards
* On Web: https://caniuse.com/webgl2 <- 93.74% over https://caniuse.com/webgl <- 98.12%

On discord people is heavily supporting going forward with this:

![Screenshot 2022-11-17 at 08 52 40](https://user-images.githubusercontent.com/570848/202387917-dbd7d22f-1869-4303-b417-607b2c1770c3.png)

So for example, if we have the layout of the attributes and uniform beforehand, we can construct a single vertex buffer and a single uniform buffer with everything. Then upload it once to the GPU, and then execute really small commands selecting memory areas in those buffers to do small render batches.
This should improve the amount of batches we can do per fram substantially.
[...]

PR #1139, opened 2022-12-01
[WIP] Completely rewritten AG, new it is much faster, vertex handling, uniform storage, support buffering batches in BatchBuilder2D to upload a single buffer + lots of other optimizations

soywiz:
This hugely improves performance of the new UIButton since it can buffer background and text.

Now we can mix multiple shaders without the performance penaly of a full batch.

Related https://github.com/korlibs/korge/issues/1116


soywiz:
Done incrementally already in the `main` branch

Issue #541, opened 2022-03-24
Include custom symbol resolver to JNA: glGetProcAddress Win32 DirectGL


soywiz:
* https://github.com/korlibs/jna/commit/6a700ace0d74528a7b14b03913dacc675097c852

We can try to patch it at runtime using javassist, and try to make a PR to the official JNA repository

This should allow to define Direct Mapping of OpenGL in all the cases including windows that require calling glGetProcAddress to get most of the GL methods, but since it uses the normal symbol resolver GetProcAdress on win32, we cannot use it.

PR #1140, opened 2022-12-02
DirectGL on Win32. Hugely reduces OpenGL overhead on windows JVM

soywiz:
Fixes https://github.com/korlibs/korge/issues/541

Needed changes for this to work:
* https://github.com/java-native-access/jna/compare/master...korlibs:jna:master
* https://github.com/java-native-access/jna/pull/1490

We should make a PR in JNA and see if we can get it directly to central, so we don't have to embed JNA into kmem

See https://github.com/java-native-access/jna/blob/master/www/DirectMapping.md


soywiz:
This functionality will be available in JNA 5.13.0 when released, so I'm pausing this PR to not include binaries here.

Please check: https://github.com/java-native-access/jna/pull/1490

soywiz:
Superseeded by https://github.com/korlibs/korge/pull/1195

PR #1148, opened 2022-12-08
Modify KorgeAndroidView.kt

MYKIM95:
- add @JvmOverloads (to use custom class in xml)
- change coroutineScope to findViewTreeLifecycleOwner().lifecycleScope


soywiz:
Thanks!

PR #1158, opened 2022-12-21
Add isActive check to animator

BioRyajenka:
Use case:
if there is already animation running, and it wasn't finished (isActive), don't do anything. Otherwise re-create animation.


soywiz:
Thanks!

PR #1159, opened 2022-12-21
Fix transparency for FSprites

### LINKED ISSUE #965: Problems with blending in FSprites
Hello,
after updating korge to 3.0 I noticed transparent images stopped to render properly with FSprites. The issue seems to be related to premultiplied images.

korge 2.0 / korge 3.0:
<p float="left">
  
  
</p>

If I comment [this line](https://github.com/korlibs/korge/blob/2b0cba9c76441171b82f0135362f4ba472cf4453/korge/src/commonMain/kotlin/com/soywiz/korge/view/fast/FSprites.kt#L274) in FSprites, it starts to work as expected.

Don't understand the details about this function though 
image

BioRyajenka:
Honestly I am not quite sure what this line does, but without it transparency work as expected.
Fixes #965

### DIFF
--- a/korge/src/commonMain/kotlin/com/soywiz/korge/view/fast/FSprites.kt
+++ b/korge/src/commonMain/kotlin/com/soywiz/korge/view/fast/FSprites.kt
@@ -267,7 +267,6 @@ open class FSprites(val maxSize: Int) {
                 SET(out, out * v_Col)
                 IF(out["a"] le 0f.lit) { DISCARD() }
                 //SET(out, vec4(1f.lit, 0f.lit, 0f.lit, 1f.lit))
-                BatchBuilder2D.DO_OUTPUT_FROM(this, out)
             }, name = "FSprites$maxTexs")
         }
 


soywiz:
@BioRyajenka https://github.com/korlibs/korge/blob/3bd7481d850f4a35ddd51cb9cc418b71d1f23918/korge/src/commonMain/kotlin/com/soywiz/korge/render/BatchBuilder2D.kt#L744-L751

Basically, input and output of shaders are expected to use colors with premultiplied alpha. The thing is that when rendering to the final output buffer, it was with straight alpha, though in most cases final render buffer alpha will be 1.0 always. I wanted to use an intermediate texture to simplify shaders and always use premultiplied for input and output, but that work was not completed

soywiz:
@BioRyajenka thanks! I have done the full cleanup now in #1162 . I have also updated tests with `./gradlew jvmTestFix`. Please check with your code if the premultiplied alpha issue is fixed with that PR too.

PR #1163, opened 2022-12-25
Call gl.beforeDoRender in AGOpengl

BioRyajenka:
Not quite sure how to create test for that, will be happy if you hint


soywiz:
Wow. Thanks for figuring it out ^^! going to make another PR adding tests for this and I'll share it with you

PR #1182, opened 2022-12-29
Go back to Kotlin 1.7.21

soywiz:


soywiz:
So people can still use next version with older versions of Kotlin for now

PR #1198, opened 2023-01-19
Fixes DirectGL on win32 and enable it, since it was not enabled

soywiz:


soywiz:
Before:

before-win32-directgl

After:

after-win32-directgl

PR #1233, opened 2023-01-29
Update with the ability to disable ligatures for TtfFont.

Kietyo:


korge/src/jvmMain/kotlin/com/soywiz/korge/testing/KorgeScreenshotTester.kt
```diff
@@ -80,9 +80,10 @@ class KorgeScreenshotTester(
     suspend fun recordGolden(
         view: View,
         goldenName: String,
-        settingOverride: KorgeScreenshotValidationSettings = defaultValidationSettings
+        settingOverride: KorgeScreenshotValidationSettings = defaultValidationSettings,
+        includeBackgroundColor: Boolean = true
     ) {
-        val bitmap = view.renderToBitmap(views)
+        val bitmap = view.renderToBitmap(views, includeBackground = includeBackgroundColor)
```
Kietyo, line 86: @soywiz - I believe there is a bug with `includeBackground`, as you can see from the image below. The red background was not actually included.

soywiz:
Can this be a non-screenshot test?
we can for example list the glyph IDs. With ligatures there will be likely two less glyphs emitted.

Kietyo:
Done.

korim/src/commonTest/kotlin/com/soywiz/korim/font/TTfTest.kt
```diff
@@ -18,30 +17,75 @@ class TTfTest {
...
+        val glyphs = mutableListOf<BaseTtfFont.Glyph>()
+
+        while (reader.hasMore) {
+            val c = reader.peek().codePoint
+            val g = ttfFontWithoutLigatures.getGlyphByReader(reader, c)!!
+            glyphs.add(g)
+        }
+
+        // The string parses to {'1', '/', '4'}
+        assertEquals(glyphs.size, 3)
```
soywiz, line 89: Since this code is duplicated, maybe we can do something to improve it?

What if we add a function inside the TtfFont with a signature like:

```kotin
fun getGlyphsByReader(reader: WStringReader, out: MutableList<Glyph> = arrayListOf()): List<Glyph>
```

Then we do the test like this (and in addition to the number of glyphs we actually ensure that the glyphs emitted are the ones expected):

```kotlin
val ttfFontWithoutLigatures = resourcesVfs["font_atkinson/AtkinsonHyperlegible-Bold.ttf"].readTtfFont(enableLigatures = false)

assertEquals(listOf(...), ttfFontWithoutLigatures.getGlyphsByReader(WStringReader("1/4")).map { it.index })
```

Kietyo: Done.

korim/src/commonTest/kotlin/com/soywiz/korim/font/TTfTest.kt
```diff
@@ -53,39 +52,29 @@ class TTfTest {
...
-            val c = reader.peek().codePoint
-            val g = ttfFontWithLigatures.getGlyphByReader(reader, c)!!
-            glyphs.add(g)
-        }
-
         // The string parses to {'¼'}
-        assertEquals(glyphs.size, 1)
+        assertContentEquals(
+            font.getGlyphsByReader(reader).map { it.index },
+            listOf(104)
```
soywiz, line 63: https://kotlinlang.org/api/latest/kotlin.test/kotlin.test/assert-equals.html

The expected value is the first argument, that might produce an inverted error if it fails. Can you reorder the arguments?

korge/src/jvmTest/kotlin/com/soywiz/korge/test/TtfFontTest.kt
```diff
@@ -0,0 +1,30 @@
...
+            centerOnStage()
+        }
+        text("without ligatures: 41/41", font = ttfFontWithoutLigatures, textSize = 40.0) {
+            centerOnStage()
+            alignTopToBottomOf(t1)
+        }
+
+        it.recordGolden(this, "text")
+
+        it.endTest()
```
soywiz, line 28: I'm a bit worried about this test. Checking the image, it seems that the antialiasing of the text is not working properly. I'm afraid that images with antialiasing or vectors with curves will likely have to be updated while we improve stuff. That PNG has a 14.1KB in size and it is already compressed, so it cannot be compressed in git and deltas won't work because a single small change will produce a totally differenet binary.

Can you disable this test and remove the PNG image, or make it as small as possible? putting in the texts only `1/4`?

Kietyo: Done

korim/src/commonTest/kotlin/com/soywiz/korim/font/TTfTest.kt
Kietyo: Done

soywiz:
Thanks!

korge/src/jvmMain/kotlin/com/soywiz/korge/testing/KorgeScreenshotTester.kt
soywiz: Lets create an issue with that

PR #1244, opened 2023-01-31
Korge multiplatform testing library.

Kietyo:


soywiz:
Do you need specific functionality related to multiplatform tests? Is it about the syntax, or missing functionality in kotlin-test? Have you tried [kotest](https://kotest.io/)?

Kietyo:
I didn't know about kotest. It seems pretty hardcore. Are we allowed to use
it in tests when developing Korlibs?

Kietyo:
My gripe with kotlin test is mainly with the syntax. Having the expected
value be the first param sucks, so i built my own thing based of the truth
library.

soywiz:
I see.

kotest seems to be supported on all the targets korlibs is using: https://github.com/kotest/kotest/blob/c2880e97401b1555c616ecc82e59105177acaa88/buildSrc/src/main/kotlin/kotest-native-conventions.gradle.kts#L10-L29
Now K/N retrocompatibility seems better, so might be a possibility to add kotest.

I have added a discussion here: https://github.com/korlibs/korge/discussions/1247

TLDR;

Initially I tried to reduce the dependencies to the very minimum. Initially there was not even kotlinx.coroutines dependency. Actually it didn't existed when korlibs started. And when it existed, K/N retrocompatibility was not a thing, and library releases (kotlinx.coroutines) took a lot of time to reach after the new compiler version was out, and we faced compiler bugs, regressions and performance issues continously so updating was required, so I always did an extra effort to reduce the dependencies to the very minimum to be able to compile for all the targets korlibs required and bump versions as fast as possible.

soywiz:
@Kietyo I'm checking all the pending PRs. So after what we wrote do you want to publish this? If so, could we rename it to something that is not `kortest` since it is really similar to `kotest`? For example, could it be `ktruth`? Is it intended to be published to central, or just to be used in the korlibs themselves?

Kietyo:
Yeah I'd like to publish, mainly so I could use this when I'm developing for Korge. I think people may find it useful if published to central, but up to you if you have the time.

soywiz:
Thanks!

PR #1269, opened 2023-02-03
Since golden references tested on the JVM might produce different results on different platforms,

soywiz:
Let's run JVM tests on windows, linux and mac independently too, so we can catch those before devs do.

@Kietyo 

Related PR: https://github.com/korlibs/korge/pull/1263


Kietyo:
Aren't git actions headless though? So the tests don't actually work in git?

soywiz:
> Aren't git actions headless though? So the tests don't actually work in git?

There are e2e tests that are working already. It was doing tricks to emulate a screen or fix opengl there. I didnt expect this was going to fail, but I cannot merge it until fixed.

In any case I added an offscreen way to initiate opengl. Maybe we can use that for the golden tests.
And figure out why macos build is crashing

PR #1278, opened 2023-02-06
[WIP] Support dynamic libraries in common code. Won't work because mingw linker and K/N hack

soywiz:
With really dirty hacks and inefficient other than with K/N

* K/N Hack: calling CPointer<Function> indirectly it is not supported. So here we redeclare the invoke functions in the same package 😧 <--- windows linker seems to fail with this hack
* Other targets Hacks: expect typealias must have the same generic arity, so we cannot mark KPointer as jna.Pointer and thus the extra hierarchy 🤕


soywiz:
Windows linker seems to fail with the K/N hack, so in the end this is not usable :/

soywiz:
Disabled the common invocation code in https://github.com/korlibs/korge/pull/1285
Closing for now

Issue #438, opened 2022-06-13
Use ALSA on linux instead of OpenAL


soywiz:
* [x] K/N and JVM ALSA implementation
* [x] Use ALSA on JVM and remove libAL.so files from the .jar
* [x] Use Waveout and Coreaudio on the JVM for Windows and Mac to remove all the embedded openal libraries

soywiz:
Done

PR #1279, opened 2023-02-06
Make common ALSA backend for Linux

soywiz:


soywiz:
More work on https://github.com/korlibs/korge/issues/438

PR #1314, opened 2023-02-11
Try to debug linux egl/mesa with envs

soywiz:
Tried docker with ubuntu server, vmware with ubuntu server images, wsl2.

The same with a window machine: tried docker with nanoserver/server and vmware with windows server 2022 minimum to try to reproduce a very small environment and was not able to reproduce context errors on windows/linux/mac.

In the case of linux I was not able to reproduce this error, except when DISPLAY/xvfb was not set. So let's try to use environment variables to figure out what's going on?

* https://docs.mesa3d.org/egl.html#environment-variables
* https://docs.mesa3d.org/envvars.html#libgl-environment-variables

### DIFF
--- a/.github/workflows/TEST.yml
+++ b/.github/workflows/TEST.yml
@@ -88,7 +88,7 @@ jobs:
     - { name: Grant execute permission for gradlew, run: chmod +x gradlew }
     - { name: Start gradle, run: ./gradlew }
     - { name: Building classes, run: ./gradlew compileTestKotlinJvm }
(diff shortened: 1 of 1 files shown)


soywiz:
Tried docker with ubuntu server, vmware with ubuntu server images, wsl2.
Was not able to reproduce this error, except when DISPLAY/xvfb was not set. So let's try to use environment variables to figure out what's going on?

* https://docs.mesa3d.org/egl.html#environment-variables
* https://docs.mesa3d.org/envvars.html#libgl-environment-variables

PR #1326, opened 2023-02-13
ASE: Support retrieving hidden layers, storing original ASE index, and MapSubject for ktruth.

Kietyo:
This supports my use case.

In my workflow, I have a background layer that is shared by a bunch of different icons.

Each layer on top are the icons themselves. I hide the icons when I develop new icons. I noticed when I hide the icon, the ASE decoder drops those layers from processing, which is annoying.

image

I also noticed that the original ASE index was getting lost due to this code here:
https://github.com/korlibs/korge/blob/31ee9785bbad665abd24fc10dec8e4e533dcdd94/korim/src/commonMain/kotlin/com/soywiz/korim/format/ASE.kt#L667

So I decided to store it separately.


korim/src/commonMain/kotlin/com/soywiz/korim/format/ASE.kt
```diff
@@ -39,6 +31,10 @@ import com.soywiz.korma.geom.RectangleInt
 import com.soywiz.korma.geom.slice.*
 import com.soywiz.krypto.encoding.hex
 
+// If this is true, only processes visible layers from the ASE file.
+// Otherwise, will process invisible layers as well.
+var ImageDecodingProps.onlyReadVisibleLayers: Boolean by extraProperty { true }
```
jobe-m, line 36: This is much more elegant than the getExtra() properties approach for "layers", "disableSlicing" and "useSlicePosition". Those are used in ParallaxDataView when reading from aseprite file.
Maybe I can align that later :)

jobe-m (approved):
LGTM!

soywiz (approved):
LGTM too 👍

.gitignore
```diff
@@ -17,3 +17,12 @@ out.txt
...
 *.hprof
+/korvi/build/
+/luak/build/
+/kbox2d/build/
+/korge-box2d/build/
+/korge-dragonbones/build/
+/korge-fleks/build/
+/korge-spine/build/
+/korge-swf/build/
+/korge-testing/build/
```
soywiz, line 28: I guess you should remove those folders yourself, because they are not longer in the repo since they were extracted to kproject

Kietyo: Done

Kietyo:
I messed up this pull request when attempting to merge, gonna create another pull request.

PR #1346, opened 2023-02-17
Proposal: Common kmem function invoke for JVM and K/N

soywiz:
This is not mergeable since this would require a patch in the Kotlin Compiler. But this should serve as a proposal of usage.

* Issue: https://youtrack.jetbrains.com/issue/KT-56446
* Kotlin Compiler proposed implementation here: https://github.com/JetBrains/kotlin/pull/5097

```bash
./gradlew :kmem:jvmTest :kmem:macosArm64Test
```


soywiz:
The proposal hack was rejected. We will have to wait for a proper implementation. Let's reopen when that's possible

Issue #1347, opened 2023-02-17
Triggers Not Recognized on Xbox One Controller


ManApart:
### Description

I no longer seem able to detect a right trigger pull on my standard xbox one controller. Instead, it is triggered by pressing the X button.
```
gamepad?.get(GameButton.RIGHT_TRIGGER)
```
and
```
val trigger = gamepad?.get(GameButton.RIGHT_TRIGGER) ?: 0.0
```
Both fire when pressing X, but not when pulling the right trigger. The controller works fine with other games, and all other inputs seem to be coming through.

### Additional Details

More testing seems to indicate that X is interpreted as right trigger, and B is interpreted as left trigger
It looks like `getButtonIndex(GameButton.XBOX_X)` and `getButtonIndex(GameButton.RIGHT_TRIGGER)` both return `14`. 
`getAxeIndex(GameButton.RIGHT_TRIGGER)` returns 5, but `info.getRawAxe(getAxeIndex(button))` returns 0.0 and that raw axes is blank when I inspect it, despite pulling the trigger... :thinking:

### Workaround

I was able to use the following to replicate the original behavior.`val trigger = gamepad?.getRawPressureButton(gamepad.mapping.getButtonIndex(GameButton.RIGHT_TRIGGER)) ?: 0.0`

ManApart:
In case it's relevant at all, here's a [code sample](https://github.com/ManApart/vex-lwjgl/blob/97c6d7092626e353d912ac8476770c544ad27c44/src/commonMain/kotlin/ui/level/Player.kt#L151) of where I ran into the issue / what my workaround was.

PR #1349, opened 2023-02-19
Fixes XInput mapping

soywiz:
Fixes https://github.com/korlibs/korge/issues/1347


ManApart (commented):
Thanks for such a speedy fix!

PR #1355, opened 2023-02-21
Common Win32 gamepad API + test

soywiz:
Using `KStructure` to be able to make common code for JVM and K/N. Still function invoking has to be done for K/N and the JVM separately for now.
With the commonization and the use of interfaces, we can also test this now without intrumentational mocks.


soywiz:
This PR is an example that could benefit from this: https://github.com/JetBrains/kotlin/pull/5097

PR #1395, opened 2023-03-08
Bump Kotlin to 1.8.20-RC

soywiz:


soywiz:
`runJs` stopped working here

soywiz:
There is a huge performance regression in the JS target here. Not sure, but might be related to the new `protoOf` thing

PR #1399, opened 2023-03-10
Feature/rename package with type alias

ygdrasil-io:
Original topics https://github.com/korlibs/korge/discussions/1249
I try and improved my KSP plugin on the kbignum project.

So far, this is partialy working, but, may be this will not be posible to create alias on extension, for example on the following code :

```
package korlibs.bignumber

// Big Integer
val Long.bi get() = BigInt(this)
val Int.bi get() = BigInt(this)
val String.bi get() = BigInt(this)
fun String.bi(radix: Int) = BigInt(this, radix)

// Big Number
val Double.bn get() = BigNum("$this")
val Long.bn get() = BigNum(this.bi, 0)
val Int.bn get() = BigNum(this.bi, 0)
val String.bn get() = BigNum(this)
```
  
The generator create this :
```
package com.soywiz.kbignum

@Deprecated("use new package korlibs.bignumber instead")
val Long.bi
 get() = korlibs.bignumber.bi
@Deprecated("use new package korlibs.bignumber instead")
val Int.bi
 get() = korlibs.bignumber.bi
@Deprecated("use new package korlibs.bignumber instead")
val String.bi
 get() = korlibs.bignumber.bi
@Deprecated("use new package korlibs.bignumber instead")
typealias bi = korlibs.bignumber.bi
@Deprecated("use new package korlibs.bignumber instead")
val Double.bn
 get() = korlibs.bignumber.bn
@Deprecated("use new package korlibs.bignumber instead")
val Long.bn
 get() = korlibs.bignumber.bn
@Deprecated("use new package korlibs.bignumber instead")
val Int.bn
 get() = korlibs.bignumber.bn
@Deprecated("use new package korlibs.bignumber instead")
val String.bn
 get() = korlibs.bignumber.bn
```

As this is an extension, we cant use the package as reference.

So, there are multiple thing that we can do
- Use the processor to generate the alias, and copy manually the extensions (we need to maintain the duplicated code until removal support of old package)
- Ignore extension
- not supporting the old package

also, i did not try yet to create alias on function.

What do you think @soywiz ?


soywiz:
Nice! This should be temporal maybe for the RC so people can bump to it and replace automatically. Then we should just remove the aliases and go clean.

ygdrasil-io:
In this case, may be only duplicate extensions used on the sandbox, there is a good of chance that the one used by other on their project.

PR #1412, opened 2023-03-14
Generate SDF for the default TTF font on the fly

soywiz:


soywiz:
This should improve text quality on retina displays

Issue #1154, opened 2022-12-17
Mipmap levels


BioRyajenka:
Hello, 

what is the best place to add mipmap levels to korge engine? 

I tried it [here](https://github.com/korlibs/korge/commit/871d3bf9e1cb17efd6dc79fe9fd3301d9ba347b0), but it doesn't work. Probably because korge creates AG.Texture somewhere under the hood, and I tried to create them from outside (with `baseMipmapLevel`, `maxMipmapLevel` fields set to the desired values initially).

Why I need levels to be adjustable: 
1. I try to do scaling with my own tilemap implementation. Scaling some sprites with `scale < .5f` gives visual artifacts.
2. To beat the artifacts I added mipmap, and it worked. But added seams to tiles.
3. To beat seams I manually adjusted GL_TEXTURE_MAX_LEVEL to the desired level and it worked. 

But I don't know how to add it to the engine, any help appreciated.

soywiz:
Regarding to the 2 bits comment: that enum might have more entries in the future.
But the code you edited doesn't exist anymore, so there is no need to pack bits there in any case anymore.

In the main branch, now all the AG opengl code happens in this file https://github.com/korlibs/korge/blob/a2aa6af912a0850b00c8f1f586ea0fda2984f13b/korgw/src/commonMain/kotlin/com/soywiz/korag/gl/AGOpengl.kt

You can search for `gl.generateMipmap` and `gl.texParameteri` to identify the best place to put your code there.

BioRyajenka:
>that enum might have more entries in the future
Yes thats true. Even though its not a big problem here (if enum will grow one can just change serialization/deserialization without any consequences), my code still have a lot of other smells. So its more for a question, not for PR yet.

**Actually the question is more is the following:**
What is the best place to integrate this functionality? From the perspective of engine architecture and API. 

Initially I added it to AG.Texture and hoped user will specify it. It was bad idea and didn't work obviously.

Another option is to add it to Bitmap class but I am also don't think its a good place. But maybe it is, considering Bitmap already have mipmaps boolean? 

Wdyt?

soywiz:
Indeed it is not something related to a Bitmap, so putting that properly as part of a Bitmap itself it wouldn't feel right. But in KorGE, textures are automatically handled and created from Bitmaps. So we need a mechanism to specify that, and ideally without complicating things too much or having to add extra parameters everywhere.

[...]

PR #1435, opened 2023-03-15
Make base & max mipmap levels specifiable

BioRyajenka:
Should resolve https://github.com/korlibs/korge/issues/1154


korgw/src/commonMain/kotlin/com/soywiz/korag/AGObjects.kt
```diff
@@ -80,6 +80,8 @@ class AGTexture(
     private val logger = Logger("AGTexture")
     var isFbo: Boolean = false
     var requestMipmaps: Boolean = false
+    var baseMipmapLevel: Int? = null
+    var maxMipmapLevel: Int? = null
```
BioRyajenka, line 84: Default value is null because if you don't specify level, GL generates some value on its own

soywiz:
Thanks!

PR #1441, opened 2023-03-19
Package renamed

ygdrasil-io:


ygdrasil-io:
Most of the work is donne by the script "rename-package.main.kts" (require kotlinc and git on a classpath to run it, working only on unix like system).

soywiz:
I would rename `package` clauses now first, but wouldn't change artifact location/names for now. In fact I wouldn't change artifact location/names in 4.0 for several reasons: do it in several steps, smoother transition for consumers, central publishing adjustments among others. Also e2e tests are failing.
If we are going to rename root folders, we should properly set the project name to the old artifact name so the publishing keeps working as before.

soywiz:
e2e test buildSrc is using korim 3.3 for checking screenshots. Those files shouldnt be updated

ygdrasil-io:
@soywiz 
```
e: Compilation failed: The C:\Users\runneradmin\.konan\dependencies\llvm-11.1.0-windows-x64-essentials/bin/clang++ command returned non-zero exit code: -1073741795.
output:
LLVM ERROR: out of memory
```

On windows, any idea what is wrong ?

soywiz:
Memory management for CI in this case is pretty complicated. Since we have the gradle daemon (JVM process), the compiler daemon (JVM process) and a LLVM linker (native process) that are racing for memory in a 8gb of RAM instance where windows is already using some resources.

Here is an additional explanation: https://github.com/korlibs/korge/blob/3a76dd04a9aabace803828b961adf4645df9a5d5/gradle.properties#L17-L26

ygdrasil-io:
@soywiz Ok, but there will surely be a Git patch to cover the edge cases. Give me a few day and i will push this.

soywiz:
thanks!

PR #1473, opened 2023-03-26
Adds operators to Frequency

lmvortex4k:
Based on TimeSpan


klock/src/commonTest/kotlin/korlibs/time/FrequencyTest.kt
```diff
@@ -9,4 +9,16 @@ class FrequencyTest {
         assertEquals(100.milliseconds, 10.timesPerSecond.timeSpan)
         assertEquals(10.hz, 100.milliseconds.hz)
     }
+
+    @Test
+    fun `test frequency operators work as expected`() {
```
soywiz, line 14: This might not work, since on JS and Android IIRC this might not be valid. Let's wait for CI

lmvortex4k: Done.

PR #1486, opened 2023-03-29
Make registerStageComponent faster

BioRyajenka:
Hello,

the motivation behind PR is to speed up `registerStageComponent` in projects with large amount of views. In my project it was 8.5% of CPU time before this change.

I understand these changes may conflict with the codestyle of the engine, will be happy to fix that if you point me.


korge/src/commonMain/kotlin/korlibs/korge/component/StageComponent.kt
```diff
@@ -36,26 +36,28 @@ fun <T : View> T.onAttachDetach(onAttach: Views.(T) -> Unit = {}, onDetach: View
...
         componentsInStagePrev.clear()
         componentsInStagePrev += componentsInStageCur
         componentsInStageCur.clear()
 
-        val stagedViews = getAllDescendantViews(stage, tempViews)
-
-        stagedViews.fastForEach { view ->
-            if (view.hasExtra(__VIEW_STAGE_COMPONENT_NAME)) {
+        viewsToTrack.forEach { view ->
+            if (view.hasExtra(__VIEW_STAGE_COMPONENT_NAME) && view.hasAncestor(stage)) {
```
BioRyajenka, line 60: `view.hasAncestor(stage)` to check the view is still on stage.
I don't like this part 🤔 Any ideas?

korge/src/commonMain/kotlin/korlibs/korge/component/StageComponent.kt
```diff
@@ -36,26 +36,28 @@ fun <T : View> T.onAttachDetach(onAttach: Views.(T) -> Unit = {}, onDetach: View
     return this
 }
 
+private val viewsToTrack = linkedSetOf<View>()
```
soywiz, line 39: This should not be done. We might have multiple korge instances. We cannot have globals. You can either add it to the Views instance as a property, or as an extra propery. Ideally extra property as this is an external component. You can retrieve the instance once at the beginning of the `registerStageComponent` function

soywiz (changes requested):
Globals should not be used

korge/src/commonMain/kotlin/korlibs/korge/component/StageComponent.kt
```diff
@@ -39,23 +39,25 @@ fun <T : View> T.onAttachDetach(onAttach: Views.(T) -> Unit = {}, onDetach: View
...
     val EXTRA_ID = "Views.registerStageComponent"
-    if (views.getExtra(EXTRA_ID) == true) return
-    views.setExtra(EXTRA_ID, true)
+
+    val firstRun = !views.hasExtra(EXTRA_ID)
+    val viewsToTrack = views.extraCache(EXTRA_ID) { linkedSetOf<View>() }
+    viewsToTrack += view
+
+    if (!firstRun) return
+
```
soywiz, line 50: I would suggest:

```kotlin
private var Views.viewsToTrack by Extra.Property<LinkedSet<View>> { null }

val firstRun = views.viewsToTrack == null
if (firstRun) {
  views.viewsToTrack = linkedSetOf()
}
...
```

that way it is typed and you don't have to use hasExtra, etc. hasExtra, setExtra, getExtra

korge/src/commonMain/kotlin/korlibs/korge/component/StageComponent.kt
soywiz: @BioRyajenka https://github.com/korlibs/korge/blob/90bc4c320eca3c5598fb58f7519366ea6a438fd1/korge/src/commonMain/kotlin/korlibs/korge/view/View.kt#L524-L525

soywiz: The `stage` reference is cached in each view, so that property won't iterate upwards.
It was an optimization in KorGE 4.0.0-alpha-1: https://github.com/korlibs/korge/commit/cdc63b4cdbac5a8a688595337d6f6427b83e0888

BioRyajenka: What I check here, is that view is assigned to the stage. When its assigned then `view.hasAncestor(stage)` is true.

Yes stage is cached but I don't understand the other part: `property won't iterate upwards`. What do you mean by that? 

Thanks.

BioRyajenka: Ohhh I see. I didn't see you propagate cache value downside as well. I will change the implementation then

soywiz (approved):

soywiz (approved):

PR #1487, opened 2023-03-29
Cache View local bounds

BioRyajenka:
Following discussion https://github.com/korlibs/korge/discussions/1476


korge/src/commonMain/kotlin/korlibs/korge/view/Container.kt
```diff
@@ -446,6 +448,7 @@ open class Container(
         old.parent = null
         new.invalidate()
         old.index = -1
+        invalidateLocalBounds()
```
soywiz, line 451: What happens if we invalidate our current view first, and then we move the view to another parent that has cached bounds? We should add a test for that. And here I guess it wouldn't invalidate parent after changing the parent.

soywiz:
I added a test covering those cases, and it seems to work. Merged already.

PR #1501, opened 2023-04-01
Add Turkish language to Klock Locales

tekinalper:


soywiz:
Thanks!

Checked:
* https://en.wikipedia.org/wiki/Turkish_months
* https://en.wikipedia.org/wiki/Names_of_the_days_of_the_week#:~:text=yetin%C3%A7%20k%C3%BCn-,Turkish,-Pazar%20%5B%E2%98%894%5D
* https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes#:~:text=tsn-,Turkish,-tr

And looks good to me 👍

PR #1508, opened 2023-04-04
Remove experimental wrapped klock variants

soywiz:
Should we remove them? Have they still use? Does kotlinx-serialization support inline classes already?


soywiz:
Since nobody complained, I'm going to remove them.

PR #1512, opened 2023-04-06
Optimize windows builds by disabling emulated-tls

soywiz:


soywiz:
Looking at a disassembler, this didn't have effect. I can see `__emutls_get_address` for simple getters in release builds. And that makes for example the 10k sample go at 15fps. We should figure out how to properly fix this

soywiz:
When removing emulatedtñs on 1.8.10 & 1.8.20, we get the following error.

Reproduced here: https://github.com/korlibs/korge/actions/runs/4627802492/jobs/8186159041?pr=1512#step:11:170

```
e: C:\Users\runneradmin\.konan\dependencies\llvm-11.1.0-windows-x64-essentials/bin/clang++ invocation reported errors


> Task :kbignum:linkDebugTestMingwX64 FAILED
The C:\Users\runneradmin\.konan\dependencies\llvm-11.1.0-windows-x64-essentials/bin/clang++ command returned non-zero exit code: 1.
output:
lld-link: error: undefined symbol: std::__once_callable
>>> referenced by D:\a\korge\korge\kbignum\src\commonTest\kotlin\korlibs\bignumber\internal\ProgressionUtilTest.kt:16
>>>               C:\Users\RUNNER~1\AppData\Local\Temp\konan_temp5187263238033016403\result.o:(kotlin::initObjectPool())

lld-link: error: undefined symbol: std::__once_call
>>> referenced by D:\a\korge\korge\kbignum\src\commonTest\kotlin\korlibs\bignumber\internal\ProgressionUtilTest.kt:16
>>>               C:\Users\RUNNER~1\AppData\Local\Temp\konan_temp5187263238033016403\result.o:(kotlin::initObjectPool())
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

When building an executable the same error is found in `Kotlin_initRuntimeIfNeeded`. So we need to report this to make it usable, and then asume that for the first release of korge 4.0 we will have ~ one order of magnitude regression over linux/mac

soywiz:
https://stackoverflow.com/questions/31617034/stdfuture-and-clang-with-stdlib-libstdc#:~:text=6%20more%20comments-,3%20Answers,-Sorted%20by%3A

soywiz:
Reported here: https://youtrack.jetbrains.com/issue/KT-56395/Kotlin-1.8.10-stdoncecallable-error-on-K-N-windows

PR #1517, opened 2023-04-10
WIP Feature/metal backend next iteration

ygdrasil-io:
Issues to fix :
- RenderContext context crash because textures are not yet supported (Renderer03 on playground)
- Normalized attributes are not well define on vertex descriptor, so they are not usable (Renderer03 on playground)
- Rendering wih multiple buffer is working, but not when sending all the vertex data on one buffer  (Renderer01 on playground)

With multiple buffer
image
Same data with one buffer
image

Of course, lot of code to clean before leaving draft status.


korgw/src/commonMain/kotlin/korlibs/graphics/AGObjects.kt
```diff
@@ -39,6 +41,7 @@ class AGBuffer : AGObject() {
     // @TODO: This will allow to upload chunks of uniform buffers for example.
     // glBufferData & glBufferSubData
     fun upload(data: ByteArray, offset: Int = 0, length: Int = data.size - offset): AGBuffer = upload(Int8Buffer(data, offset, length).buffer)
+    fun upload(data: UByteArray): AGBuffer = upload(Uint8Buffer(data).buffer)
```
soywiz, line 44: Missing the `offset` and `length` to be consistent with the other signatures. For simplicity, since UByteArray is a wrapper of ByteArray, maybe you can do something like `upload(data.asByteArray(), offset, length)` and reuse the ByteArray method

settings.gradle
```diff
@@ -29,7 +29,7 @@ include(":ktruth")
 include(":korge-gradle-plugin")
 include(":korge-reload-agent")
 include(":korge-sandbox")
-if (System.getenv("ENABLE_OSX_METAL_PLAYGROUND") == "true") {
+if (enableMetalPlayground) {
```
soywiz, line 32: Alternatively we can use an environment called DISABLE, and updated the `.github/workflows` to disable it, to ensure it is not published to maven central. Or disable publishing in the build gradle

soywiz:
Nice! As long as tests are passing, the artifact is not published, and OpenGL is still used by default with maybe metal enabled with a feature flag, we can merge it when you feel like it 👍

korgw/src/commonMain/kotlin/korlibs/graphics/AGObjects.kt
ygdrasil-io: Not sure to keep this implementation, but if i do, il will update the code.

settings.gradle
ygdrasil-io: Using environment variable is not convenient on IntelliJ, i will switch of the value on the gradle.properties before leaving the draft status.

ygdrasil-io:
> Nice! As long as tests are passing, the artifact is not published, and OpenGL is still used by default with maybe metal enabled with a feature flag, we can merge it when you feel like it even if not finished 👍

I will try to fix the current issues before leaving draft status.

.run/run sandbox.run.xml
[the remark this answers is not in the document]
soywiz: Weren't run configurations available in the `.idea` folder? What's this `.run` folder BTW?

settings.gradle
soywiz: Well, this is for configuring, not for running. But in any case, my main concern here is: not publishing/installing and ideally not building this module in CI. Maybe the quickest solution is `if (System.getenv("CI") != "true") {`. That way on contributor's computers the module would be available directly, and no extra configuration required in workflows.

.run/run sandbox.run.xml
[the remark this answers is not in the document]
ygdrasil-io: Not if you share it explicitly. As `.idea` folder should be consider as local file and not be commited, you can share configuration with "store as project file".
image

This is the idiomatic way on IntelliJ to help other or remember how to run specific task.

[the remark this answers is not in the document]
soywiz: Thanks for the info! Tried it and looks like a good idea. Let's upload it. I have created other with runJvm here: https://github.com/korlibs/korge/pull/1521

soywiz:
BTW. Regarding to uniform blocks:

https://github.com/korlibs/korge/blob/d1fe53fcd8da349daacbc9ccbc816b4d88d6906c/korgw/src/commonMain/kotlin/korlibs/graphics/shader/UniformBlock.kt#L48

In the case of OpenGL I had to place each element in multiple of 256, to comply alignment requirements on OpenGL to use uniform blocks.
Is it possible in metal to specify the offset of each field/the size of the struct?

In the case of OpenGL we have something like:

```
// Buffer is like
// PROP1,PROP2,PADDING256 ... PROP1,PROP2,PADDING256 ...
uploadUniformBufferToGPU()
setUniformBlockPointer(addressMultipleOf256)
draw()
setUniformBlockPointer(addressMultipleOf256)
draw()
```

In the case of metal, since we have a pointer to the start of the buffer, and probably we pass an index referring to structs in that buffer, we have to ensure the generated buffer size matches the struct size. 
https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf
It seems that we cannot use unions to force a specific size for structs. So maybe when using metal we should either see if we can get a pointer based on the index and an arbitrary size for each element. Or repack stuff before uploading the buffer.

soywiz:
Hey @ygdrasil-io to prevent this from getting too outdated, let me grab this PR, update it so it compiles and tests are passed or disabled and merge it. I have done that already on WASM code to avoid code getting too outdated.

Issue #696, opened 2021-07-17
Update JVM and Native renderers to offload rendering to a separate thread/worker, and perform vsync in that thread while processing events on the main thread while waiting vsync


soywiz:
This will fix https://github.com/korlibs/korge/issues/591

Issue #591, opened 2021-07-04
IO / HTTP request overhead on KorGE default dispatcher


Kesanov:
Issuing any http request on JVM has a ~20s delay.

For example, downloading a `3Kb` file with `UrlVfs` takes around ~25 seconds:
```
UrlVfs("https://raw.githubusercontent.com/korlibs/korio/master/README.md").readString()
```
Where 20s is waiting for the first byte!
Wrapping the request in a `runBlockingNoJs` seems to remove all of this overhead.

PR #1541, opened 2023-04-18
[WIP] Separated thread for updating and rendering on the JVM

soywiz:
This should: improve drawing performance, hugely improve effective performance on I/O since events won't have to be executed once every 16ms, but as soon as possible. Implemented only on the JVM for now.

Some work on https://github.com/korlibs/korge/issues/525
Some work on https://github.com/korlibs/korge/issues/696
Some work on https://github.com/korlibs/korge/issues/591


soywiz:
To avoid interferring with the 4.0 release, I'm going to defer this to 4.1, since this is highly experimental. Other targets like K/N can follow a similar approach too.

soywiz:
Superseded by https://github.com/korlibs/korge/pull/1962

PR #1562, opened 2023-05-01
runJs: correct MIME type for wasm files

smaugfm:
Currently wasm files are served with `application/octet-stream` by the builtin `staticHttpServer` task used in `runJs` which causes browser error, i.e. with `WebAssembly.instantiateStreaming`.
I understand that wasm is not supported by Korge yet, but it can still be used in source-sets independent from Korge, i.e. to offload heavy game computations.


soywiz:
Thanks!

smaugfm:
Any chance for this to come up in a future 4.0.0 release candidate?

soywiz:
This will be included in 4.0.0 final. That will be released next week.

PR #1565, opened 2023-05-02
Atlas texture generation for *.atlas folder/files, fixes catalog generation and resource processing on all the targets & adds mechanism to add a custom resource processors in the build.gradle

soywiz:


jobe-m:
This looks very promising. Will it be possible to use Aseprite files as input for this atlas generation at gradle time?
If not, maybe we can generate an atlas file out from Aseprite meta data.

soywiz:
@jobe-m 

We had some problems with the classpath because we included the current korge libraries and compiled the jvm target of the current project to detect new plugins. To avoid some issues I disabled that for now and we dont have access to korim/korio anymore in the plugin, so loading ase files with korim in the plugin is not possible anymore.

We can try to enable it again in >= 4.1 to support loading psd, qoi and ase files too for atlas generation in the plugin by using korim.

Feel free to open a feature request for that 👍

jobe-m:
I created #1566

PR #1567, opened 2023-05-03
runJs: correct MIME type for JavaScript modules

smaugfm:
Similar to this [one](https://github.com/korlibs/korge/pull/1562), but now for the [commonly](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types) used JavaScript `*.mjs` module files.

### DIFF
--- a/buildSrc/src/main/kotlin/korlibs/korge/gradle/util/HttpServerTools.kt
+++ b/buildSrc/src/main/kotlin/korlibs/korge/gradle/util/HttpServerTools.kt
@@ -175,7 +175,7 @@ fun File.miniMimeType() = when (this.extension.toLowerCase()) {
     "svg" -> "image/svg+xml"
     "mp3" -> "audio/mpeg"
     "wasm" -> "application/wasm"
-    "mjs" -> "text/javascript"
+    "js", "mjs" -> "text/javascript"
     else -> if (this.exists()) Files.probeContentType(this.toPath()) ?: "application/octet-stream" else "text/plain"
 }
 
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/korlibs/korge/gradle/util/HttpServerTools.kt
```diff
@@ -175,6 +175,7 @@ fun File.miniMimeType() = when (this.extension.toLowerCase()) {
     "svg" -> "image/svg+xml"
     "mp3" -> "audio/mpeg"
     "wasm" -> "application/wasm"
+    "mjs" -> "text/javascript"
```
soywiz, line 178: Maybe `”js”, “mjs” ->`?

smaugfm: `js` is correctly recognized by `Files.probeContentType`. But I can add it for completeness' sake.

soywiz: Yes, add it too. In the end we cannot control directly which files are detected. Probably PNG and HTML are already detected but are there also. Maybe newer JVM versions are already providing the right mime-types. But since we are providing others too, let's provide all the ones that we might want to serve so this is complete.

smaugfm: Done👌

soywiz:
Thanks!

Issue #1586, opened 2023-05-10
Bring back the non-inline wrapper classes


dalewking:
4.0 deleted the Wrapped version of the classes (e.g. WDateTime), but they are still needed for iOS. See https://github.com/korlibs/korge/pull/1508#issuecomment-1542639209

dalewking:
In the end it doesn't really matter for us, since we are going to move to kotlinx-datetime, using Klock only for formatting.

PR #1604, opened 2023-05-15
Revert "Remove experimental wrapped klock variants (#1508)"

soywiz:
This reverts commit 982d8666ddb9fe91a75801a2b4c0f9f96132128b.

Would fix https://github.com/korlibs/korge/issues/1586

Let's figure out if people is still using it again.


soywiz:
Hey folks 👋 https://github.com/korlibs-archive/klock/issues/108

@lamba92 @RomeuG are you still using this? I can add the wrapped versions again if it makes sense.

RomeuG:
Hi @soywiz! Not anymore, in my case. Thank you for the heads up.

soywiz:
In the end I'm going to merge it, just in case some people still uses this but didn't see this RFC so they can bump without losing functionality.

dalewking:
My team is definitely still using this and it actually is preventing us from updating to latest version. BUT we will be migrating away from Klock over the next month so not that big of a deal.

soywiz:
No worries. It was clear to me. But since I'm not sure how much people might be in the same situation, and it is something that can be done easily I think it might benefit other people. Eventually the idea is to use kotlinx.datetime and wrap around that. But in the meantime this can be a solution for those people.

PR #1614, opened 2023-05-23
Prepare to update to Kotlin 1.9.0-Beta (not updated just yet but make required changes)

soywiz:


soywiz:
https://youtrack.jetbrains.com/issue/KT-58837/Illegal-char-at-index-0-org.jetbrains.kotlinxatomicfu-cinterop-interop-CTypeDefinitions

PR #1683, opened 2023-06-11
Make ImageAnimationView & ImageDataView Anchorable. Adds PixelAnchorable

soywiz:


jobe-m:
Does pixelAnchor mean I can set the anchor point in pixels? Should come in handy, I think :)

soywiz:
Yes. To have an stable anchoring in the case the children change. Used it here: https://github.com/korlibs/korge-starter-kit-dungeon

PR #1716, opened 2023-06-20
Try to bump to Kotlin 1.9.0-RC

soywiz:


soywiz:
In 1.8.22 tests were compiled but not properly executed. So for example:

https://youtrack.jetbrains.com/issue/KT-58681/K-Wasm-division-remainder-has-a-wrong-sign

We might need to wait for 1.9.20, or disable the wasm target.

soywiz:
Superseeded by https://github.com/korlibs/korge/pull/1777

Issue #1062, opened 2022-10-22
Revisit Angle working based on typically inverted Y axis


soywiz:
https://discord.com/channels/728582275884908604/801658049244430377/1030771780652113980

PR #1717, opened 2023-06-21
Supports Orientation, Point.orientation, Point.polar & Angle with Y going down (inverted up vector)

soywiz:
We will need to do the same with Angle.between & Point.angleTo. Maybe alternatively we could provide a UP vector as reference, so it is even more generic.

Fixes https://github.com/korlibs/korge/issues/1062

@Kietyo


soywiz:
WIP: Let's add code documentation, website documentation & update Angle.between and Point.angleTo

Issue #669, opened 2020-11-08
Translate ordinals (1st, 2nd...) into several languages [de, es, fr, it, ja, ko, nb, nl, pt, ru, sv, uk, zh]


soywiz:

Bantolomeus:
Hi @soywiz I wanted to push a branch for the german ordinals, but I am not allowed. Do you need to grant me rights?

soywiz:
> Hi @soywiz I wanted to push a branch for the german ordinals, but I am not allowed. Do you need to grant me rights?

Hey @Bantolomeus you have to fork this repository, push the changes in your fork and make a PR with the changes

PR #1724, opened 2023-06-25
De ordinals

Bantolomeus:
I appreciate a review of the code. It feels not right to write down all 32 cases as enum, but german ordinals are partly irregular up to 19. I could not come up with a better solution.

### DIFF
--- a/klock/src/commonMain/kotlin/korlibs/time/locale/de.kt
+++ b/klock/src/commonMain/kotlin/korlibs/time/locale/de.kt
@@ -8,7 +8,7 @@ val KlockLocale.Companion.german get() = GermanKlockLocale
 
 open class GermanKlockLocale : KlockLocale() {
 
-    companion object {
+    companion object : GermanKlockLocale() {
         private val NUM_TO_ORDINAL: Array<String> = arrayOf(
             "index zero is invalid",
             "erste",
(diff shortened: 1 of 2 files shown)


Bantolomeus:
This PR should solve the german part of https://github.com/korlibs/korge/issues/669.

soywiz:
@Bantolomeus thanks for the PR!

Looking at the code I would suggest a change in how it is done. Let me elaborate on that:
* By using the enum, it is needed to replace some characters to support in `getOrdinalInGerman` the `ä`, `ö`, `ü`. The enum adds an overhead
* The code ends using `GermanOrdinal.values()` which generates a new array and then iterate over all the elements.
* The enum elements are not used as literals anywhere, so I would suggest a Map here. To avoid creating a new class, with all the elements, that is normally more heavyweight in some cases.

So my proposal is to do something like:

```
companion object {
  private val NUM_TO_ORD: Array<String> = arrayOf(
      "", // index 0 (invalid)
      // ...
      "zwölfte", // index 12 (no need to do the replace `oe` to `ö`)
      // ...
  )
  private val ORD_TO_NUM: Map<String, Int> = ORD_TO_NUM.mapIndexed { index, s -> s to index }.toMap()
}
```

Then accessing the array can be done with `ORD_TO_NUM.getOrNull(index) ?: error("Invalid numeral")` and the opposite with `ORD_TO_NUM[key.toLowerCase()]`.

Bantolomeus:
@soywiz I hope deleting `companion object : GermanKlockLocale()` from the class does not break anything elsewhere, but I was not allowed to define two companion objects in the class.
Thanks for the earlier elaboration.

klock/src/commonMain/kotlin/korlibs/time/locale/de.kt
```diff
@@ -1,12 +1,58 @@
...
 import korlibs.time.KlockLocale
+import korlibs.time.KlockLocaleContext
+import korlibs.time.DayOfWeek
 
 val KlockLocale.Companion.german get() = GermanKlockLocale
 
 open class GermanKlockLocale : KlockLocale() {
-	companion object : GermanKlockLocale()
+
+    companion object {
```
soywiz, line 11:
```suggestion
    companion object : GermanKlockLocale() {
```

klock/src/commonMain/kotlin/korlibs/time/locale/de.kt
```diff
@@ -32,4 +78,46 @@ open class GermanKlockLocale : KlockLocale() {
 
 	override val formatTimeMedium = format("HH:mm:ss")
 	override val formatTimeShort = format("HH:mm")
+
+    enum class GermanOrdinal(val value: Int) {
```
soywiz, line 82: This enum should be removed

soywiz:
Can add some tests covering these German locale ordinals?

Bantolomeus:
sure

klock/src/commonMain/kotlin/korlibs/time/locale/de.kt
Bantolomeus: My bad, wanted to delete it.

klock/src/commonMain/kotlin/korlibs/time/locale/de.kt
Bantolomeus: done

Bantolomeus:
I will add the test either later today or tomorrow.

Bantolomeus:
I could not come up with a good test for the ordinals so I copied your english one. But that is kinda useless, since it does exactly the same as the german ordinal code since I can not add a suffix to the german ordinals. 
Please tell me if the test shall stay or if I should remove it.
And I changed the index 0 to "nullste" since you did it that way in english, to stay consistent.
If you have not further concerns you can merge the PR from my end.

soywiz:
Looks good to me.
Don't remember the english version, but since it is not going to be accessed the first element could be an empty string or a null. Then it could be checked as `if (number !in 1..31) outofboundsexception`.
In any case, looks good to me as it is now 👍. Thanks for the PR @Bantolomeus !

Bantolomeus:
Thanks for Guiding me through it. It was my pleasure and my very first open source contribution =D

Bantolomeus:
The native windows Test failed, but it do not understand what the error is, something with coroutines, but nothing concrete. Was the Test flaky?

soywiz:
It is flaky indeed. Nothing to do with this code. I have re-run it, and likely will be merged now.

PR #1726, opened 2023-06-26
add option for undecorated AWT window

cmhulbert:
Adds a flag for creating an undecorated window. Currently, only meaningful for AWT windows.

### DIFF
--- a/korge/src/commonMain/kotlin/korlibs/korge/Korge.kt
+++ b/korge/src/commonMain/kotlin/korlibs/korge/Korge.kt
@@ -59,7 +59,8 @@ suspend fun Korge(
     debug: Boolean = false,
     trace: Boolean = false,
     context: Any? = null,
-    fullscreen: Boolean? = null,
+    @DeprecatedParameter("Use WindowConfig instead") fullscreen: Boolean = false,
+    windowConfig: WindowConfig = WindowConfig(),
     blocking: Boolean = true,
     gameId: String = Korge.DEFAULT_GAME_ID,
     settingsFolder: String? = null,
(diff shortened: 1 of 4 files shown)


soywiz:
Nice. Thanks!

@cmhulbert 
What would you think about creating a `WindowConfig` class like 
`data class WindowConfig(val undecorated: Boolean, val transparent: Boolean, val allowMaximize, val allowMinimize..., val allowFullscreen: Boolean)` and then passing it instead of the undecorated to the korge method?

korge/src/commonMain/kotlin/korlibs/korge/Korge.kt
```diff
@@ -60,6 +60,7 @@ suspend fun Korge(
     trace: Boolean = false,
     context: Any? = null,
     fullscreen: Boolean? = null,
+    undecorated: Boolean? = null,
```
soywiz, line 63: Can you create a class called `WindowConfig`, initially with `data class WindowConfig(val undecorated: Boolean = false)`, so we can add there other stuff without polluting there method more.

Full proposal in the comment.

soywiz (changes requested):

cmhulbert:
Yes! Sorry, haven't looked at this again yet, but I think that's a good idea. I hope to have some time to push some more changes tonight.

cmhulbert:
So, I put something similar to the changes you suggested. I'm open to suggestions for improvements. A couple things I noticed while doing this:
- There already is a `GameWindowCreationConfig` data class. should we just add to this, instead of a new `WindowConfig`?
- There is also a `GameWindow.configure` method which accepts `title`, `size`, `icon`, and `bgColor` as well. Should those all be put in the `WindowConfig` options?
- should `title` be in the `WindowConfig`?
- you suggested `transparent`. Is that separate from the window background color being specified as `Colors.TRANSPARENT`?

soywiz:
> So, I put something similar to the changes you suggested. I'm open to suggestions for improvements. A couple things I noticed while doing this:
> 
> * There already is a `GameWindowCreationConfig` data class. should we just add to this, instead of a new `WindowConfig`?

Yes, just checked, and this is in the common source set, so we should be able to use it when creating KorGE:
https://github.com/korlibs/korge/blob/d530d16e95471f2c0e9b17e0b3f116b7c0fb8e3b/korgw/src/commonMain/kotlin/korlibs/render/GameWindow.kt#L32-L40
I guess we should unify it there.

> * There is also a `GameWindow.configure` method which accepts `title`, `size`, `icon`, and `bgColor` as well. Should those all be put in the `WindowConfig` options?

Probably yes.

> * should `title` be in the `WindowConfig`?

Maybe yes too. But keeping the old properties as `@Deprecated`. Since it is a data class, maybe we can do something like:

```kotlin
data class Widow...Config...(val title: String = "...Game...")
fun Korge(
  title: String? = null,
  config: Window...Config...,
)... {
  // ...
  val config = config.copy(title = title ?: config.title, ...)
  // ...
}
```

So we can use the deprecated methods, but omiting them use the new WindowConfig way.

> * you suggested `transparent`. Is that separate from the window background color being specified as `Colors.TRANSPARENT`?

I would say yes. Though we can detect `Colors.TRANSPARENT` if it is provided in the `GameWindowConfig`

soywiz:
I have backported it

PR #1747, opened 2023-07-01
fix: korge crashing after too many sounds are played on linux

### LINKED ISSUE #1728: Crash when attempting to play sounds repeatedly on JVM with ALSA
In my game, I am playing a walking sound for each step the character takes. It may be interrupted, for example if the player changes direction mid step. Every time a sound is played, I get the following stack trace:
```
Exception in thread "Thread-115" java.lang.InterruptedException: sleep interrupted
	at java.base/java.lang.Thread.sleep(Native Method)
	at java.base/java.lang.Thread.sleep(Thread.java:339)
	at korlibs.time.internal.KlockInternal.sleep-82NKd7o(KlockInternal.kt:15)
	at korlibs.time.SleepKt.blockingSleep-N3vl5Ow(Sleep.kt:11)
	at korlibs.audio.sound.backends.ALSAPlatformAudioOutput$start$6.invoke(ALSA.kt:96)
	at korlibs.audio.sound.backends.ALSAPlatformAudioOutput$start$6.invoke(ALSA.kt:120)
	at korlibs.datastructure.thread.NativeThread.thread$lambda$0(NativeThread.kt:4)
	at java.base/java.lang.Thread.run(Thread.java:834)
```
After a handful of these (maybe 30 or so?) I get the following error, and the korge exits
```
ALSA lib pulse.c:242:(pulse_connect) PulseAudio: Unable to connect: Connection terminated

java: pcm.c:1560: snd_pcm_writei: Assertion `pcm' failed.
Can't initialize ALSA
```

Here is a reduced example that duplicates the issue for me; `STEP` can be any `Sound`:
```kt
suspend fun main() {
    while (true) {
        val soundChannel = STEP.play()
        soundChannel.await()
    }
}
```

Anecdotally, It also seems that the interval between the sounds get subtly longer as more and more iterations of the sound play. I haven't measured this, but it sounds like it 

I'm running on Pop!Os 22.04, building against the JVM target. I'm currently using a manually built Korge version tracking the most recent `main` branch commit (6060a27bb70aab4dcf02ccaa0ffbd5761643b23e when last tested)

cmhulbert:
Code mostly provided by @soywiz, with minor modifications of my own. Thank you!
Closes #1728


soywiz:
Thanks!

Issue #1748, opened 2023-07-01
Blurry Text in uiScrollable


cmhulbert:
When using text in a `uiScrollable` text is much blurrier:

![image](https://github.com/korlibs/korge/assets/1859669/8b8b78af-4892-415d-82ff-1a99a4472a28)

minimal script to reproduce: 
```kt
suspend fun main() = Korge(
        windowSize = Size(400, 800),
        virtualSize = Size(100, 200)
    ) {
    uiVerticalStack(100f) {
        uiScrollable(size = Size(100, 100)) {
            textBlock(align = TextAlignment.MIDDLE_CENTER, size = Size(100, 100)) {
                name = "blurry"
                text = RichTextData("test", font = DefaultTtfFontAsBitmap, textSize = 16.0f)
                blendMode = BlendMode.NORMAL
                fill = Colors.WHITE
            }
        }
        textBlock(align = TextAlignment.MIDDLE_CENTER, size = Size(100, 100)) {
            name = "sharp"
            text = RichTextData("test", font = DefaultTtfFontAsBitmap, textSize = 16.0f)
            blendMode = BlendMode.NORMAL
            fill = Colors.WHITE
        }
    }
}
```

cmhulbert:
Actually, it's a bit more involved I guess. It looks like it has to do with the default font. Maybe when nested in the `uiScrollable`, the other font is not being respected? Here is a side by side. Top row is in a scrollable, left column has `font = DefaultTtfFontAsBitmap` and right column has default font:


![image](https://github.com/korlibs/korge/assets/1859669/75ddcd7c-b849-4d58-bd13-80e1af155f67)

```kt
suspend fun main() = Korge(
        windowSize = Size(800, 800),
        virtualSize = Size(200, 200)
    ) {
    uiHorizontalStack {
        uiVerticalStack {
            uiScrollable(size = Size(100, 100)) {
                textBlock(align = TextAlignment.MIDDLE_CENTER, size = Size(100, 100)) {
                    name = "blurry"
                    text = RichTextData("in scrollable ttfAsbitmap", font = DefaultTtfFontAsBitmap)
                    fill = Colors.WHITE
                }
            }
            textBlock(align = TextAlignment.MIDDLE_CENTER, size = Size(100, 100)) {
                name = "sharp"
                text = RichTextData("alone ttfAsbitmap", font = DefaultTtfFontAsBitmap)
                fill = Colors.WHITE
            }
        }
        uiVerticalStack {
[...]

PR #1760, opened 2023-07-04
Fix blurry DefaultTtfFontAsBitmap text in CachedContainer when virtual size < window size

eaboll:
Adds `CachedContainer.expensiveScaling: Boolean?`, which enables scaling of the content based on the actual window size instead of just the virtual size. Set to `true` by default. 

This fixes part of issue #1748. The default non-bitmap font is still blurry regardless of whether or not it's in a CachedContainer.

In the sandbox:

image

### DIFF
--- a/korge/src/commonMain/kotlin/korlibs/korge/view/CachedContainer.kt
+++ b/korge/src/commonMain/kotlin/korlibs/korge/view/CachedContainer.kt
@@ -108,7 +108,7 @@ open class CachedContainer(
                 //ctx.ag.clear(Colors.TRANSPARENT, clearColor = true)
                 ctx.setViewMatrixTemp(globalMatrixInv
                     .translated(-lbounds.x, -lbounds.y)
-                    .scaled(windowLocalRatio)
+                    .scaled(renderScale * windowLocalRatio.scaleX, renderScale * windowLocalRatio.scaleY)
                 ) {
                     super.renderInternal(ctx)
                 }
@@ -120,7 +120,7 @@ open class CachedContainer(
                 cache.tex,
                 m = globalMatrix
                     .pretranslated(lbounds.x, lbounds.y)
-                    .prescaled(1.0 / windowLocalRatio.scaleX, 1.0 / windowLocalRatio.scaleY)
+                    .prescaled(1.0 / renderScale / windowLocalRatio.scaleX, 1.0 / renderScale / windowLocalRatio.scaleY)
                 ,
                 colorMul = renderColorMul,
                 blendMode = blendMode,
--- a/korge/src/commonMain/kotlin/korlibs/korge/view/CachedContainer.kt
+++ b/korge/src/commonMain/kotlin/korlibs/korge/view/CachedContainer.kt
@@ -36,7 +36,9 @@ open class FixedSizeCachedContainer(
 
 open class CachedContainer(
     @property:ViewProperty
-    var cache: Boolean = true
+    var cache: Boolean = true,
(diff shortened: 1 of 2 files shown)


soywiz:
Awesome. Thanks for implementing it!

I have noticed a small tweak that could improve it further if you can change it

korge/src/commonMain/kotlin/korlibs/korge/view/CachedContainer.kt
```diff
@@ -36,7 +36,9 @@ open class FixedSizeCachedContainer(
 
 open class CachedContainer(
     @property:ViewProperty
-    var cache: Boolean = true
+    var cache: Boolean = true,
+    @property:ViewProperty
+    var expensiveScaling: Boolean = true,
```
soywiz, line 41: Can we have `expensiveScaling: Boolean? = null` ?

When constructing KorGE we have a `quality` parameter. Can we use the quality parameter to determine the value when not specified?

Here's the quality:
https://github.com/korlibs/korge/blob/b9213f1adc4d66c2148fca9363ca520874a1a696/korgw/src/commonMain/kotlin/korlibs/render/GameWindow.kt#L480

soywiz: This PR https://github.com/korlibs/korge/pull/1762/files will be needed, so it is possible to access tue GameWindow quality inside the RenderContext

eaboll: > Can we have `expensiveScaling: Boolean? = null` ?

Excellent idea, implemented! https://github.com/korlibs/korge/pull/1760/commits/59cb48385fc87c92190146f4d759a8187d3cdd5e

korge/src/commonMain/kotlin/korlibs/korge/view/CachedContainer.kt
```diff
@@ -84,16 +84,21 @@ open class CachedContainer(
...
+        val renderScale: Float = when (ctx.quality) {
             GameWindow.Quality.PERFORMANCE -> 1f
             else -> ctx.devicePixelRatio
         }
         //val renderScale = 1.0
 
+        val doExpensiveScaling = expensiveScaling ?: when(ctx.quality) {
+            GameWindow.Quality.PERFORMANCE -> false
+            else -> true
+        }
```
soywiz, line 96: Do you think that by adding some properties the Quality enum could we avid somehow the when so the code is simpler?

korge/src/commonMain/kotlin/korlibs/korge/view/CachedContainer.kt
```diff
@@ -36,7 +36,9 @@ open class FixedSizeCachedContainer(
 
 open class CachedContainer(
     @property:ViewProperty
-    var cache: Boolean = true
+    var cache: Boolean = true,
+    @property:ViewProperty
+    var expensiveScaling: Boolean? = null,
```
soywiz, line 41: I'm not sure about this name. Could we pass a `quality: Quality` enum, or rename it? Like `hidpiAware: Boolean?` or other names? People might ask: In which way is this expensive? What scaling mean in this context?

soywiz:
Thanks @eaboll for implementing this!

To avoid this getting outdated, I'm going to merge it and make a few adjustments.
I'm going to change `CachedContainer.expensiveScaling` to `CachedContainer.quality` and going to adjust Quality to be accessible from outside, and to simplify the when.

PR #1770, opened 2023-07-05
Fix/alsa wait

cmhulbert:
The previous commit did fix ALSA not being interruptible, but I notice that it also doesn't cleanly end on it own (i.e. without being interrupted). Specifically, `availableSamples` was not implemented, so `wait()` always returned immediately. This was not noticed, since `stop()` called `drain()` which waits for the written audio to end. This seeks to solve that by:

- properly updating `availableSamples` so `wait()` works properly
- change `stop()` to call `snd_pcm_drop()` instead of `snd_pcm_drain()` to immediately drop all remaining frames, instead of waiting for them to finish
- calling `wait()` then `stop()` when a sound is done
- calling only `stop()` when a sound is cancelled
- implement `snd_pcm_delay()` to get the number of `frames` remaining to be heard after `snd_pcm_writei` to determine how long to wait before writing the next sound (or returning, if done).

### DIFF
--- a/korau/src/commonMain/kotlin/korlibs/audio/sound/backends/ALSA.kt
+++ b/korau/src/commonMain/kotlin/korlibs/audio/sound/backends/ALSA.kt
@@ -33,6 +33,7 @@ class ALSAPlatformAudioOutput(
     coroutineContext: CoroutineContext,
     frequency: Int,
 ) : PlatformAudioOutput(coroutineContext, frequency) {
+    override var availableSamples: Int = 0
     val channels = 2
     private val lock = Lock()
     val sdeque = AudioSamplesDeque(channels)
@@ -45,7 +46,8 @@ class ALSAPlatformAudioOutput(
     }
 
     override suspend fun add(samples: AudioSamples, offset: Int, size: Int) {
-        if (!ASound2.initialized) return super.add(samples, offset, size)
+        if (size == 0) return
+        if (!ASound2.initialized) return super.add(samples, offset, samples.totalSamples)
         //if (!running) start()
 
         if (!running) delay(10.milliseconds)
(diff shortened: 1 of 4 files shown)


korau/src/commonMain/kotlin/korlibs/audio/sound/backends/ALSA.kt
```diff
@@ -45,15 +46,17 @@ class ALSAPlatformAudioOutput(
...
         //if (!running) start()
 
         if (!running) delay(10.milliseconds)
 
         while (running && lock { sdeque.availableRead > 4 * 1024 }) {
             delay(10.milliseconds)
         }
-        lock { sdeque.write(samples, offset, size) }
+        availableSamples += samples.totalSamples
+        lock { sdeque.write(samples, offset, samples.totalSamples) }
```
soywiz, line 59: might make sense to put `availableSamples +=` inside the lock to ensure it is thread safe?

korau/src/commonMain/kotlin/korlibs/audio/sound/backends/ALSA.kt
```diff
@@ -127,16 +131,24 @@ class ALSAPlatformAudioOutput(
...
                     if (result == -ASound2.EPIPE) {
                         ASound2.snd_pcm_prepare(pcm)
+                    } else {
+                        while (running && ASound2.snd_pcm_delay(pcm) > frames) {
+                            blockingSleep(10.milliseconds)
+                        }
                     }
+
+
+                    availableSamples -= samples.totalSamples
```
soywiz, line 146: ditto, lock?

korau/src/commonMain/kotlin/korlibs/audio/sound/backends/ALSA.kt
```diff
@@ -127,16 +131,24 @@ class ALSAPlatformAudioOutput(
...
                 }
             } finally {
                 //println("COMPLETED: $pcm")
                 thread = null
-//                stop()
+                availableSamples = 0
```
soywiz, line 151: lock?

cmhulbert:
All modification of `availableSamples` are now locked

soywiz:
Thanks!

PR #1778, opened 2023-07-09
Successfully debug incorrect colors when rendering grayscale jpg images.

Kietyo:
The issue was due to the `.cloneIfRequired(type = type)` line.

image

I'm creating this pull request to share my findings with you.

Hoping that you can take care of it from here since I'm afraid to mess up the color type stuff since I don't really know how that stuff works.

After cloning, you can run the `runJvmJvmMain` gradle task to see that the colors are now correct:

image


soywiz:
Thanks! I will investigate

soywiz:
@Kietyo I have checked this.

First of all:
Unless you found something strange, if the tests are passing, I believe it should be mergeable. If we find that something stopped working, we can add a regression test later or now and fix it. There are already tests covering image decoding, so please don't worry too much about it if the tests are passing :)

I have checked the test result: https://github.com/korlibs/korge/actions/runs/5498927502/jobs/10020758616?pr=1778#step:11:2662
and it seems that a test is trying to read an absolue file in this path `C:/Users/kietm/GitHub/ImageCropper/src/commonMain/resources/converted.jpg`
Can you add it to the resources and load it from `resourcesVfs` like the rest of the tests?

Only that test seem to be failing, so maybe just by fixing that is enough.

Extra points if you can make [this change](https://github.com/korlibs/korge/pull/1778/files#diff-352a4ed6705bb7a917575c5a819be8660ac68e9e7eebd706777e9eb1bc3432bb) in a separate commit.

Issue #1788, opened 2023-07-11
Convert Closeable -> AutoCloseable


soywiz:
In Kotlin 1.8, `AutoCloseable` was included. A first step could be making `Closeable` extend `AutoCloseable`.
And then typealiasing it.
Expending the interface is okay, typealiasing it will be a binary-breaking change, and should be done only in major version changes.

ygdrasil-io:
`Context2d` has a conflict with `AutoCloseable`. It inherit from `VectorBuilder` wich contains a close method.
The `close` method will be deprecated in flavor of `closePath` on 4.1.0 and remove on 4.2.0

PR #1791, opened 2023-07-11
Make Closeable extend AutoCloseable as a first step to migrating to it

soywiz:
Some work on https://github.com/korlibs/korge/issues/1788


ygdrasil-io:
next version here https://github.com/korlibs/korge/pull/1834

PR #1821, opened 2023-08-04
Improve constructors of often single-initialization classes to accept…

Kietyo:
… Number to provide easier to use API.

### DIFF
--- a/korge/src/commonMain/kotlin/korlibs/korge/render/RenderContext2DExt.kt
+++ b/korge/src/commonMain/kotlin/korlibs/korge/render/RenderContext2DExt.kt
@@ -120,19 +120,19 @@ object MaterialRender {
 
 @KorgeExperimental
 fun RenderContext2D.materialRoundRect(
-    x: Float,
-    y: Float,
-    width: Float,
-    height: Float,
+    x: Number,
+    y: Number,
+    width: Number,
+    height: Number,
     color: RGBA = Colors.RED,
     radius: RectCorners = RectCorners.EMPTY,
     shadowOffset: Point = Point.ZERO,
     shadowColor: RGBA = Colors.BLACK,
-    shadowRadius: Float = 0f,
+    shadowRadius: Number = 0f,
     highlightPos: Point = Point.ZERO,
-    highlightRadius: Float = 0f,
+    highlightRadius: Number = 0f,
     highlightColor: RGBA = Colors.WHITE,
-    borderSize: Float = 0f,
+    borderSize: Number = 0f,
     borderColor: RGBA = Colors.WHITE,
     //colorMul: RGBA = Colors.WHITE,
 ) {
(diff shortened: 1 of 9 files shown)


solonovamax:
a downside of using `Number` instead of `Float`:
on the JVM, any time you want to use those classes, it will incur a performance impact due to having to constantly box/unbox the floats/etc.

alternatively, a secondary constructor could be offered that just invokes `toFloat()` on everything.

soywiz:
When creating an `inline fun` variant doing the casts IIRC, it was optimized on the JVM. On K/N debug targets it was not, and performance was pretty unusable the last time I tried. Doing it in the main constructor is likely to box. Creating an instance is already costly so maybe it is worth for those cases, and we might keep the variants or making inline methods that don't allocate.

solonovamax:
are there benchmarks to test this?
if not, that's something that could be added using kotlinx-benchmark or smth to be able to see the difference in performance it makes

also, it seems I didn't read the commit history fully lol so I didn't realize they all already are secondary constructors

tbh, what I'd recommend is rather than using `Number` is having a secondary constructor for `Long`, `Int`, and `Double`, then the base class holds a `Float`. this should avoid all boxing on the JVM because it will compile down to primitives and primitive casts.

variants for `Short` and `Byte` can also be created, but those are very rarely used.

I would go for secondary constructors wherever possible, regardless of the cost of creating an instance, as often on the JVM it can do funny escape analysis stuff and determine that the class is only ever used as a wrapper, and then optimize out the class entirely for smth equivalent to a C struct. (and whenever valhalla comes along, it'll be already ready for that)

soywiz:
I was not able to setup kotlinx.bencharmk with Kotlin/Native the last time I tried.

I believe we can do something like:

```
class MyClass(val value: Float) {
  companion object {
    inline operator fun invoke(value: Number) = MyClass(value.toFloat())
  }
}

// or:

inline fun MyClass(value: Number) = MyClass(value.toFloat())

```

That would act like a constructor in both cases, but can be inline, so the cast can be optimized by the compiler.
This was suggested because there were precissely several variants with Int, Float, Double, etc. and that was annoying, both for maintaining, for coding, for mixing numbers, like: (1.5, 1, 0) etc.

solonovamax:
> This was suggested because there were precissely several variants with Int, Float, Double, etc. and that was annoying, both for maintaining, for coding, for mixing numbers, like: (1.5, 1, 0) etc.

ah, I see

also, you could do
```kotlin
class MyClass(val value: Float) {
    constructor(value: Number) : this(value.toFloat())
}
```

the "several variants with `Int`, `Float`, `Double`", was it a different class for each, or just one class with a bunch of secondary constructors?

because having `Number` there incurs the cost of boxing, which can be pretty costly at times (less allocations is always better)

Kietyo:
Apologies I believe this got a bit out of hand. I forgot to mention my intention was just to update constructors for **UI components** (e.g SolidRect, UI dropdown, UI text field, etc) which are **often just initialized once for a scene**, and thus would not impact performance at all (I would in order of nanosecond impact).

I believe for classes such as Point, we should be much more wary of boxing and such but my intention was never to update that class as I know it's used magnitudes more than the UI components.

PR #1825, opened 2023-08-07
remove unused code

ygdrasil-io:


Kietyo:
We should be very careful about this. 
Much of the "unused" code are code for clients to use so they will only appear unused from our perspective.
The multiplatform code is especially impacted by this and the infra may actually need that code for the multiplatform builds.

ygdrasil-io:
> We should be very careful about this. Much of the "unused" code are code for clients to use so they will only appear unused from our perspective. The multiplatform code is especially impacted by this and the infra may actually need that code for the multiplatform builds.

Of course so far, i only removed old implementation and commented code. Do you want to do a review if test are ok ?

Kietyo:
LGTM

PR #1826, opened 2023-08-08
move kbignum to kproject

ygdrasil-io:
First check to see if there is any impact on e2e tests.
@Kietyo do you see any reason to keep this on the core library and not on a kproject ?


Kietyo:
IMO it is easier to develop in the mega repo since all korlibs dependencies are in one place which makes it easier to make changes across libs and not have to worry about the kproject depending on older versions of libs.

From a user POV I also prefer using maven/gradle dependencies and i imagine for most users they will not take the time to understand kproject and integrate it into their project. I also find that it is hard to integrate kproject into multi module projects since it's confusing which dependencies are being used if you use both kproject and regular dependencies { } block.

ygdrasil-io:
@Kietyo 

> IMO it is easier to develop in the mega repo since all korlibs dependencies are in one place which makes it easier to make changes across libs and not have to worry about the kproject depending on older versions of libs.

I agree that we should keep a mono repo to stuff related to korge.
Still, some projects like kbignum are not related to the engine and can have a different lyfecycle of release than korge.
What do you think if 
- we keep thing as it on version 4.0.x
- we deprecate library not related directly to korge on 4.1.X while release is as new library
- we remove them from this repo on 4.2.X 

> From a user POV I also prefer using maven/gradle dependencies and i imagine for most users they will not take the time to understand kproject and integrate it into their project. I also find that it is hard to integrate kproject into multi module projects since it's confusing which dependencies are being used if you use both kproject and regular dependencies { } block.

I agree that maven dependencies are more convenient that kproject to use.

Kietyo:
One downside though is that kbignum will no longer be included into future maven releases.

Sine it's basically only me and you maintaining this now, I would still prefer keeping it into the mega repo for ease of development.

That being said, if you feel strongly about this feel free to submit.

PR #1838, opened 2023-08-22
fix base16 xml entity

lmvortex4k:


korio/src/commonTest/kotlin/korlibs/io/serialization/xml/XmlTest.kt
```diff
@@ -153,4 +153,11 @@ class XmlTest {
             xml.take(10).toList().joinToString("\n")
         )
     }
+
+    @Test
+    fun testEntities() {
```
ygdrasil-io, line 158: Could we use a more descriptive name that clearly indicates what we're testing in this case?

korio/src/commonMain/kotlin/korlibs/io/serialization/xml/Xml.kt
```diff
@@ -253,7 +253,15 @@ data class Xml(
 				val value = r.readUntilIncluded(';') ?: ""
 				val full = "&$value"
 				when {
-					value.startsWith('#') -> append(value.substring(1, value.length - 1).toInt().toChar())
+					value.startsWith('#') -> {
```
ygdrasil-io, line 256: Perhaps this should be in a private method to adhere to the Single Responsibility Principle?

PR #1841, opened 2023-09-13
Unify modules and folders

soywiz:


solonovamax:
What was the reason for deciding to merge these modules?

I have a few projects that use kds and/or kmem, but not klock, klogger, or korau. and, I would really prefer to not have to include those in my projects if I don't need them.

soywiz:
> What was the reason for deciding to merge these modules?
> 
> I have a few projects that use kds and/or kmem, but not klock, klogger, or korau. and, I would really prefer to not have to include those in my projects if I don't need them.

Focus.

I have forked it before doing this: https://github.com/korlibs/korlibs4 . Community can contribute to that repo for those libs and keep releasing 4.x.y versions there and backport fixes and features from/to there.

solonovamax:
I see. I understand that.

is there any way that (specifically) kds and kmem could still be maintained as separate modules, or no?

soywiz:
Yes. The korlibs4 repository is the way to go for that.

PR #1828, opened 2023-08-08
Migrate to .kts for korge-gradle-plugin.

Kietyo:

PR #1852, opened 2023-09-18
Convert some build.gradle into build.gradle.kts

soywiz:
Retake on https://github.com/korlibs/korge/pull/1828


soywiz:
Let's see how much does it affect to CI

PR #1873, opened 2023-09-25
Added TypedArrays implementations mimicking JS and type-aliasing in that target

soywiz:
Backported from Korge5 prototype


soywiz:
I guess this behaviour is wrong:

Screenshot 2023-09-25 at 12 29 34

Shouldn't be a problem not having default arguments in the expected while having in the actual. This changed in 1.9.20

PR #1884, opened 2023-09-27
Go back to Kotlin 1.9.10 and disable wasm target for now

soywiz:


soywiz:
It seems to be a bug triggered in the older Kotlin version. We will stick to the beta for the 5.0.0 and will bump to the final version in 5.0.x

PR #1894, opened 2023-09-29
add SolidTriangle

rafi0:
Hi soywiz :)
I added a possibility to render a triangle using the existing mechanism of rectangles, to make use of the existing GPU pipeline.
I added one small hacky solution, that I will try to fix in the upcoming days. 
Thanks


rafi0:
i updated the implementation with mesh - there is however a problem with color - Mesh doesn't have that. Probably need to use bitmap for that?

soywiz:
We can set `Mesh.colorMul`:
https://github.com/korlibs/korge/blob/dd3901bd23fbadf6e9ce5b2fd8f5a86a31012b6b/korge/src/common/korlibs/korge/view/View.kt#L460

It is computed here renderColorMul: https://github.com/korlibs/korge/blob/dd3901bd23fbadf6e9ce5b2fd8f5a86a31012b6b/korge/src/common/korlibs/korge/view/Mesh.kt#L39C7-L39C12

Basically, in KorGE all the drawings are textured drawings. When solid colors, we have a blank white 1x1 bitmap, and then we tint it with the `View.colorMul` property

korge/src/common/korlibs/korge/view/SolidTriangle.kt
```diff
@@ -0,0 +1,25 @@
...
+  val vertices = floatArrayOf(
+    p1.x.toFloat(), p1.y.toFloat(),
+    p2.x.toFloat(), p2.y.toFloat(),
+    p3.x.toFloat(), p3.y.toFloat(),
+  )
+  val indices = Uint16Buffer(UShortArrayInt(shortArrayOf(0, 1, 2)))
+  val uvs = Float32Buffer(12)
+  return Mesh(vertices = Float32Buffer(vertices), indices = indices, uvs = uvs)
+    .addTo(this, callback)
+}
```
soywiz, line 25: I believe we should either extend Mesh or compose with `Container` + addChild.

That way we can expose `p1`, `p2`, `p3` properties in the SolidTriangle class to recompute the vertices without removing the view and creating a new one

korge/src/common/korlibs/korge/view/RectBase.kt
```diff
@@ -47,7 +47,7 @@ open class RectBase(
 	val sRight: Float get() = sLeft + bwidth
 	val sBottom: Float get() = sTop + bheight
 
-    protected val vertices = TexturedVertexArray(4, TexturedVertexArray.QUAD_INDICES)
+    protected open val vertices = TexturedVertexArray(4, TexturedVertexArray.QUAD_INDICES)
```
soywiz, line 50: I guess this can be reverted now

korge/src/common/korlibs/korge/view/SolidRect.kt
```diff
@@ -21,7 +21,7 @@ inline fun Container.solidRect(width: Int, height: Int, color: RGBA = Colors.WHI
 /**
  * A Rect [RectBase] [View] of size [widthD] and [heightD] with the initial color, [color].
  */
-class SolidRect(size: Size, color: RGBA = Colors.WHITE) : RectBase() {
+open class SolidRect(size: Size, color: RGBA = Colors.WHITE) : RectBase() {
```
soywiz, line 24: Ditto reverted

korge/src/common/korlibs/korge/view/RectBase.kt
rafi0: Done.

korge/src/common/korlibs/korge/view/SolidRect.kt
rafi0: Done.

korge/src/common/korlibs/korge/view/SolidTriangle.kt
```diff
@@ -0,0 +1,48 @@
...
+      p2.x.toFloat(), p2.y.toFloat(),
+      p3.x.toFloat(), p3.y.toFloat(),
+    )
+  ),
+  indices = Uint16Buffer(UShortArrayInt(shortArrayOf(0, 1, 2))),
+  uvs = Float32Buffer(12)
+) {
+  var p1: Point = p1
+    set(value) {
+      updatedVertices()
```
rafi0, line 24: This doesn't seem to work - updating the p1 is not updating the rendered triangle

korge/src/common/korlibs/korge/view/SolidTriangle.kt
soywiz: This commit: https://github.com/korlibs/korge/commit/09fa71e0e5cbb50aa633461bf46e460627a68e63
fixes it, right?

korge/src/common/korlibs/korge/view/SolidTriangle.kt
```diff
@@ -21,23 +21,36 @@ class SolidTriangle(
...
   var p2: Point = p2
     set(value) {
+      updateTriangle()
       updatedVertices()
       field = value
     }
   var p3: Point = p3
     set(value) {
+      updateTriangle()
       updatedVertices()
```
soywiz, line 37: These calls to `updatedVertices` might be moved to updateTriangle?

korge/src/common/korlibs/korge/view/SolidTriangle.kt
```diff
@@ -21,23 +21,36 @@ class SolidTriangle(
...
       field = value
     }
 
   init {
     colorMul = color
   }
+
+  fun updateTriangle() {
```
soywiz, line 45: Can we make this private?

soywiz:
So I would instantiate the Mesh with an empty `Float32Buffer(6)`.
Then we can call the updateTriangles in the `init { }` block.
Instead of creating a new buffer in the updateTriangles, to avoid allocations, we can mutate the float32Buffer.
Like `vertices[0] = value`, etc.

korge/src/common/korlibs/korge/view/SolidTriangle.kt
rafi0: You're right. Done.

korge/src/common/korlibs/korge/view/SolidTriangle.kt
rafi0: Done.

rafi0:
> So I would instantiate the Mesh with an empty `Float32Buffer(6)`. Then we can call the updateTriangles in the `init { }` block. Instead of creating a new buffer in the updateTriangles, to avoid allocations, we can mutate the float32Buffer. Like `vertices[0] = value`, etc.

Done.

soywiz:
LGTM! Thanks!


@rafi0 
Now we could try to add a screenshot test for this in a separate PR? 

Have a look here: https://github.com/korlibs/korge/blob/main/korge/test/jvm/korlibs/korge/view/SDFViewTest.kt

`gradlew runJvmFix` to generate the screenshot

PR #1899, opened 2023-10-01
Feature/solidtriangle screenshot test

rafi0:


soywiz:
Nice.
Can you change the three points after the screenshot assert, and do another screenshot assert to verify that changing the points is updating the triangle?

PR #1925, opened 2023-10-05
Three-argument document.open()

Bruce0203:
corrected redirection in js


korge/src/js/korlibs/render/DialogInterfaceJs.kt
```diff
@@ -11,7 +11,7 @@ import org.w3c.files.*
 class DialogInterfaceJs : DialogInterface {
 
     override suspend fun browse(url: URL) {
-        document.open(url.fullUrl)
+        document.open(url.fullUrl, "", "noopener=true")
```
soywiz, line 14: How that I notice. Shouldn't it be `window.open`?
https://developer.mozilla.org/en-US/docs/Web/API/Window/open

PR #2027, opened 2023-11-12
added piechart + screenshot test

rafi0:


rafi0:
@soywiz 
Please let me know if you like the package i created this in: korge.datapresentation
later on i will also implement line chart and a bar chart and maybe some other stuff

rafi0:
@soywiz also the failing test seems flaky - it works on my end xD (windows)

soywiz:
Sorry for the huge delay, I have been pretty busy this week and some stuff accumulated.
I believe the flaky problem is related to text rendering being slightly different on each computer. To solve that there is a PSNR parameter and a posterize parameter to allow that differences.

On the other hand, what would you think on creating a `korge-charts` repository and expose this as a kproject in the store? That would give it better visibility and the releate iteration won't need to be the same.

soywiz:
Hey @rafi0 I have backported your code here: https://github.com/korlibs/korge-charts
Feel free to make PRs there

rafi0:
> Sorry for the huge delay, I have been pretty busy this week and some stuff accumulated. I believe the flaky problem is related to text rendering being slightly different on each computer. To solve that there is a PSNR parameter and a posterize parameter to allow that differences.
> 
> On the other hand, what would you think on creating a `korge-charts` repository and expose this as a kproject in the store? That would give it better visibility and the releate iteration won't need to be the same.

No problem :) 

> Hey @rafi0 I have backported your code here: https://github.com/korlibs/korge-charts
> Feel free to make PRs there

Thanks, i will continue there

soywiz:
Great. I have also called:

The `./addlink https://github.com/korlibs/korge-charts/tree/main/korge-charts`

In this commit:
https://github.com/korlibs/store.korge.org/commit/1672598e859eb08619cec6bc4b827bb48a61e677

So it appears here and can be installed using korge intellij plugin:
https://store.korge.org/module/korge-charts/

PR #2032, opened 2023-11-15
Fix UIButton background color not updating properly

maik-mursall:
This fixes the following bug:
- The UIButton's background color doesn't update when setting `enabled` from `false` to `true` (enabling from disabled state)

**Implementation:**
- Improved and slimmed down `UIButton.updatedUIButton`

### DIFF
--- a/korge/src/common/korlibs/korge/ui/UIButton.kt
+++ b/korge/src/common/korlibs/korge/ui/UIButton.kt
@@ -102,8 +102,10 @@ open class UIButton(
     override var enabled: Boolean
         get() = super.enabled
         set(value) {
+            if (value != mouseEnabled)
+                updatedUIButton(enabled = value)
+
             mouseEnabled = value
-            updatedUIButton(enabled = value)
         }
 
     //protected val rect: NinePatchEx = ninePatch(null, width, height)
--- a/korge/src/common/korlibs/korge/ui/UIButton.kt
+++ b/korge/src/common/korlibs/korge/ui/UIButton.kt
@@ -99,6 +99,13 @@ open class UIButton(
             setInitialState()
         }
 
+    override var enabled: Boolean
+        get() = super.enabled
+        set(value) {
+            mouseEnabled = value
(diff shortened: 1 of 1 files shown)


korge/src/common/korlibs/korge/ui/UIButton.kt
```diff
@@ -99,6 +99,16 @@ open class UIButton(
...
 
+    override var enabled: Boolean
+        get() = super.enabled
+        set(value) {
+            val valueChanged = value != mouseEnabled
+            mouseEnabled = value
+
+            if (valueChanged)
+                updatedUIButton(enable = value)
+        }
```
soywiz, line 110: Why not updating UIView from:

```suggestion
open var enabled: Boolean
    get() = mouseEnabled
    set(value) {
        mouseEnabled = value
        updateState()
    }
```

to

```suggestion
open var enabled: Boolean
    get() = mouseEnabled
    set(value) {
        if (mouseEnabled == value) return
        mouseEnabled = value
        updateState()
    }
```

? so we don't need overrigint this? wouldn't that work?

Since `updateState` is calling `updatedUIButton`, and since mouseEnabled has been set, when updatedUIButton gets the enabled property is already set to the right value, right?

maik-mursall: @soywiz I would have prefered it that way as well, but we need that to be able to supply the `enable` argument for `updatedUIButton`.

If you have a nicer idea to find out, if `enabled` changed since the last update, I would be happy to see that :D

maik-mursall: And I liked your version for checking if the last value is the same within the setter method, so I changed it to yours :)

soywiz (approved):

korge/src/common/korlibs/korge/ui/UIButton.kt
soywiz: Okay. Let's keep it as this is 👍

PR #2038, opened 2023-11-24
Fix ViewStyles not passing down to children

maik-mursall:
- Fixes `ViewStyles` not being passed down to children.

#### Explanation
This Bug occurs when the parent is not an `UIView`.
One example is `SContainer`, which is only a `View` and not a `UIView`, meaning the `styles` of an `SContainer` will never pass down to their children.
Meaning you can never define one global style within your apps entrypoint.

### DIFF
--- a/korge/src/common/korlibs/korge/style/ViewStyles.kt
+++ b/korge/src/common/korlibs/korge/style/ViewStyles.kt
@@ -5,7 +5,6 @@ import korlibs.image.color.*
 import korlibs.image.font.*
 import korlibs.image.text.*
 import korlibs.io.concurrent.atomic.*
-import korlibs.korge.ui.*
 import korlibs.korge.view.*
 import kotlin.reflect.*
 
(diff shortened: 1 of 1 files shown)


soywiz:
Woops :) nice catch!

We should add a test for covering it.

https://github.com/korlibs/korge/blob/main/korge/src/common/korlibs/korge/style/ViewStyles.kt
->
https://github.com/korlibs/korge/tree/main/korge/test/common/korlibs/korge -> style package `ViewStylesTest.kt`. 

Tests for reference:
* https://github.com/korlibs/korge/blob/a1c3c0a965fc3665bd0d7fec7155c65a014242d1/korge/test/common/korlibs/korge/view/ContainerTest.kt

I guess what we could test here is:
* We create a `Container()` and and maybe a few more descendants, including an UIView
* Set styles in a parent container that is not an UIView
* `assertEquals` checking that the UIView or a descendant of it, gets the proper style inherited from the arent container that was not an UIView

What do you think?

maik-mursall:
@soywiz sounds good 👍🏻 

I will try implement some tests for this :)

I would add a test for overriding the styles as well.
Something similar to:
- Create a `Container` with a style
- Have another nested `Container` with a style as well
- Check if children within the second `Container` use the style of the second `Container`

I'll hit you up, if I encounter any problems :D

maik-mursall:
@soywiz I just added some tests :).

Please check if they're good like that and if I missed something :b.

soywiz:
Awesome! Looks good to me. Thanks!

PR #2057, opened 2023-12-02
Enable wasmJs

soywiz:
Now that kotlinx.coroutines supporting wasmJs is available on central


soywiz:
This seems to be the reason of the hang:

image

Issue #2046, opened 2023-11-26
Reintroduce Kotlin Native targets


soywiz:
For all the modules excluding KorGE (where JVM is used for desktop)
And only targets supported by kotlinx-coroutines.

jamesward:
I think this means we will see a linuxX64 target which I'd love to have especially for the IO / Compression stuff.

soywiz:
That's the idea. But I won't have too much time for this in the short term. PRs are welcome.

PR #2070, opened 2023-12-18
Re-enable K/N desktop for korlibs-time and try to test using github action matrix

soywiz:
Includes:

* `linuxX64`, `linuxArm64`
* `macosX64`, `macosArm64`
* `mingwX64`

For now only enabled on `korlibs-time`. Later it will be included to the rest of libraries up to `korge-core`.

See https://github.com/korlibs/korge/issues/2046 for details


soywiz:
Maybe: https://stackoverflow.com/questions/65384420/how-do-i-make-a-github-action-matrix-element-conditional

PR #2090, opened 2023-12-28
fix URL.resolve and PathInfo.normalize

itboy87:
## Changes in `URL.resolve`

- Added new test cases in `URLTest.testResolve`.
- Removed old query on the copy of the path.
- Added protocol for a relative URL.
  - `assertEquals("https://example2.com/one", URL.resolve("https://example.com/", "//example2.com/one"))`
- Now returns the base URL if the access URL is empty.
  - `assertEquals("https://example.com/one", URL.resolve("https://example.com/one", ""))`
- Introduced a new extension function `String.normalizeUrl`.
  - Avoids normalizing query parameters or URL fragments.
  - Keeps end slashes after normalization.

## Changes in `PathInfo.normalize`

- Added new test cases in `URLTest.testNormalize`.
- Added a parameter for `removeEndSlash`.
  - This parameter is introduced to match the standard of the web, but in filesystem paths, it's not required, and the default value is `true`.
  - `assertEquals("g", "g/".pathInfo.normalize())`
  - `assertEquals("g/", "g/".pathInfo.normalize(removeEndSlash = false))`


itboy87:
I just found that I have missed the following test case, which is failing with the this PR code: `assertEquals("https://example.com/path/file?foo", URL.resolve("https://example.com/path/file?bar", "?foo"))`
This PR code resolves it to https://example.com/path/?foo, which is not correct. I will fix this and also add this test case.

itboy87:
I have addressed the resolution of parameters or fragments. The fix should pass the following test:

```kotlin
assertEquals("https://example.com/path/file?foo", URL.resolve("https://example.com/path/file?bar", "?foo"))
```

However, our code still doesn't resolve URLs correctly for protocols like ftp: or file:, as illustrated in the following example:
```kotlin
assertEquals("file://var/log/messages", URL.resolve("file://etc/", "/var/log/messages"))
```

These types of URLs lack a host, resulting in incorrect resolution to file://etc/var/log/messages. I suspect FTP behaves similarly. URLs that don't begin with http should be handled differently. For example, if the access starts with / and the base is not an HTTP URL, it can be represented as `protocol://access`, omitting the starting / from the access part. Implementing this approach will resolve the issue. What are your thoughts on this?

Additionally, as per [RFC3986 Secion-3.1](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1) , URL schemes should be case-insensitive, producing only lowercase scheme names for consistency. Should I apply this change in this pull request as well?

itboy87:
Additionally, as per [RFC3986 Secion-3.1](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1) , URL schemes should be case-insensitive, producing only lowercase scheme names for consistency. Should I apply this change in this pull request as well?

itboy87:
@soywiz I have one more scenario to handle: if the base URL is empty or not absolute, how should it be resolved? I think we should return the access URL if it's absolute; otherwise, return an empty string or null. 
In `java.net.URL("")` or `java.net.URL("test")`, a `java.net.MalformedURLException` exception is thrown.

soywiz:
I have created two issues for covering your suggestions: 
* https://github.com/korlibs/korge/issues/2092
* https://github.com/korlibs/korge/issues/2093

Feel free to make PRs to handle those. And thanks!

itboy87:
> I have created two issues for covering your suggestions:
> 
> * [URL schemes should be case-insensitive, producing only lowercase scheme names for consistency #2092](https://github.com/korlibs/korge/issues/2092)
> * [if the base URL is empty or not absolute, how should it be resolved? I think we should return the access URL if it's absolute; otherwise, return an empty string or null. #2093](https://github.com/korlibs/korge/issues/2093)
> 
> Feel free to make PRs to handle those. And thanks!

ok Great.

PR #2097, opened 2023-12-30
Handle empty or invalid URLs when resolving, fixes #2093

### LINKED ISSUE #2093: if the base URL is empty or not absolute, how should it be resolved? I think we should return the access URL if it's absolute; otherwise, return an empty string or null.
https://github.com/korlibs/korge/pull/2090#issuecomment-1871659252
https://github.com/korlibs/korge/pull/2090#issuecomment-1871831284

itboy87:
This will fixe issue #2093.

## Changes in `URL.resolve`
* Throws `MalformedInputException` if the base URL and access URL are empty or invalid.
* Returns the access URL if the base URL is invalid but the access URL is absolute.
* Test cases added.

## New function: `URL.resolveOrNull`
Introduces a new function to avoid exceptions and return `null` when resolving URLs.


soywiz:
Great. Thanks @itboy87 !

PR #2102, opened 2023-12-31
URL schemes always lowercase fixes issue #2092

### LINKED ISSUE #2092: URL schemes should be case-insensitive, producing only lowercase scheme names for consistency
https://github.com/korlibs/korge/pull/2090#issuecomment-1871419169

itboy87:
As per [RFC3986 Secion-3.1](https://datatracker.ietf.org/doc/html/rfc3986#section-3.1) , URL schemes should be case-insensitive, producing only lowercase scheme names for consistency.

This PR closes #2092


soywiz:
Thanks!

PR #2109, opened 2024-01-01
fix CharReaderFromSyncStream read characters less than chunkSize (#2108)

### LINKED ISSUE #2108: Unexpected results with CharReaderFromSyncStream when chunk size is less than reading count
I suggest buffering up only inside the `tempStringBuilder.length < count` block after checking if more buffer is needed, rather than before. I will work on it and send a PR.

#### Test case
```kotlin
val reader = "áéíóúñ".toByteArray(UTF8).toCharReader(charset = UTF8, chunkSize = 2)
assertEquals("á,éí,óúñ", listOf(reader.read(1), reader.read(2), reader.read(10)).joinToString(","))
```

Expected : `á,éí,óúñ`
Actual   : `á,é,í`

itboy87:
There was an issue with `CharReaderFromSyncStream`. When attempting to read characters fewer than the specified `chunkSize`, it returns the maximum `chunkSize` data instead. I have set a minimum `chunkSize` limit to 8 because the decoder requires a minimum size of the maximum character size to decode. I believe the maximum character size in bytes is 4 bytes, but for safety, I used 8 bytes. Therefore, if anyone sets the chunkSize to less than 8, it will throw an exception.

* Created two new constants:
```kotlin
companion object {
       private const val DEFAULT_CHUNK_SIZE = 1024
       private const val MIN_CHUNK_SIZE = 8
}
```

* Check `chunkSize` and throw `IllegalArgumentException` exception if its less than `MIN_CHUNK_SIZE`
```kotlin
init {
  if (chunkSize < MIN_CHUNK_SIZE) throw IllegalArgumentException("chunkSize must be greater than $MIN_CHUNK_SIZE")
}
```

* Refactored buffer fill code to `bufferUp()` function. This function returns the number of bytes added to the buffer. If `tempStringBuilder` doesn't have enough characters to read, the function will attempt to read more into the buffer until the condition is met or the stream is empty.

This PR closes #2108

### DIFF
--- a/korge-core/src/korlibs/io/stream/CharReader.kt
+++ b/korge-core/src/korlibs/io/stream/CharReader.kt
@@ -13,33 +13,50 @@ fun ByteArray.toCharReader(charset: Charset, chunkSize: Int = 1024): CharReader 
 fun SyncStream.toCharReader(charset: Charset, chunkSize: Int = 1024): CharReader =
     CharReaderFromSyncStream(this, charset, chunkSize)
 
-class CharReaderFromSyncStream(val stream: SyncStream, val charset: Charset, val chunkSize: Int = 1024) : CharReader {
+class CharReaderFromSyncStream(val stream: SyncStream, val charset: Charset, val chunkSize: Int = DEFAULT_CHUNK_SIZE) : CharReader {
     private val temp = ByteArray(chunkSize)
     private val buffer = ByteArrayDeque()
     private var tempStringBuilder = StringBuilder()
 
+    init {
+        if (chunkSize < MIN_CHUNK_SIZE) throw IllegalArgumentException("chunkSize must be greater than $MIN_CHUNK_SIZE")
+    }
+
     override fun clone(): CharReader = CharReaderFromSyncStream(stream.clone(), charset, chunkSize)
 
     override fun read(out: StringBuilder, count: Int): Int {
-        while (buffer.availableRead < temp.size) {
-            val readCount = stream.read(temp)
-            if (readCount <= 0) break
-            buffer.write(temp, 0, readCount)
-        }
-
+        bufferUp()
         while (tempStringBuilder.length < count) {
             val readCount = buffer.peek(temp)
             val consumed = charset.decode(tempStringBuilder, temp, 0, readCount)
-            if (consumed <= 0) break
-            buffer.skip(consumed)
+            if (consumed <= 0) {
+                if (bufferUp() <= 0) break
+            } else {
+                buffer.skip(consumed)
+            }
         }
-
-        //println("tempStringBuilder=$tempStringBuilder")
 
         val slice = tempStringBuilder.substring(0, kotlin.math.min(count, tempStringBuilder.length))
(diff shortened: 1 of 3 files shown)


korge-core/test/korlibs/io/net/URLTest.kt
```diff
@@ -115,6 +115,18 @@ class URLTest {
...
+        assertEquals("https", url3.scheme)
+        assertEquals("username:password", url3.userInfo)
+        assertEquals("username", url3.user)
+        assertEquals("password", url3.password)
+        assertEquals("example.com", url3.host)
+        assertEquals(8443, url3.port)
+        assertEquals("/path/to/resource", url3.path)
+        assertEquals("/path/to/resource?q=1", url3.pathWithQuery)
+        assertEquals("q=1", url3.query)
+        assertEquals("https://username:password@example.com:8443/path/to/resource?q=1", url3.fullUrl)
```
soywiz, line 129: I guess this should be part of another PR

korge-core/src/korlibs/io/stream/CharReader.kt
```diff
@@ -13,33 +13,50 @@ fun ByteArray.toCharReader(charset: Charset, chunkSize: Int = 1024): CharReader
...
     CharReaderFromSyncStream(this, charset, chunkSize)
 
-class CharReaderFromSyncStream(val stream: SyncStream, val charset: Charset, val chunkSize: Int = 1024) : CharReader {
+class CharReaderFromSyncStream(val stream: SyncStream, val charset: Charset, val chunkSize: Int = DEFAULT_CHUNK_SIZE) : CharReader {
     private val temp = ByteArray(chunkSize)
     private val buffer = ByteArrayDeque()
     private var tempStringBuilder = StringBuilder()
 
+    init {
+        if (chunkSize < MIN_CHUNK_SIZE) throw IllegalArgumentException("chunkSize must be greater than $MIN_CHUNK_SIZE")
```
soywiz, line 22: Really minor an optional, but: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/require.html

soywiz (changes requested):
URLText extra test should be done in a separate PR since it looks unrelated to this

itboy87:
@soywiz ok i will update it.

korge-core/test/korlibs/io/net/URLTest.kt
itboy87: ok i will add that in PR for fixing URL `subScheme`

korge-core/src/korlibs/io/stream/CharReader.kt
itboy87: Done replaced with `require`

korge-core/test/korlibs/io/net/URLTest.kt
itboy87: Done moved to another PR.

soywiz (approved):
LGTM

soywiz:
Thanks!

Issue #2040, opened 2023-11-24
Split korge-foundation and korge-core in submodules


soywiz:
Like the old libs, but for package name. Keeping `korge-foundation` and `korge-core` as a module without code linking to each module individually.

* `korge-time`
* `korge-crypto`
* `korge-datastructure`
* ...

So everything works as before but people can get the libraries segmented to what they need.

soywiz:
Do this after merging / discarding PRs to avoid conflicts.

soywiz:
I won't have time for this in the shorterm. If someone wants to contribute here, PRs are welcome.

* `korge-datastructure` is already picked by `holloszaboakos`

If someone wants to contribute in this split with PRs, feel free to do so. You can put a comment here to avoid other people to doing the same work.

holloszaboakos:
We have a problem with circular dependency. We might have to do a bigger refactor than just separating the packages into modules

holloszaboakos:
The math and the logger modules use the korge-datastructure, therefor we have circular dependency of gradle tasks and this problem is not solvable without gradle changes

soywiz:
@holloszaboakos it should be fine:

Just add `commonMainApi(project(":korlibs-datastructure"))` in `korge-foundation/build.gradle.kts` so the math and logger packages have access to it.

> If something from other package is required into datastructure, just copy that functionality into an internal package and provide those functions/classes as internal so they are not exposed.

> When `logger` and `math` are split into modules, they can just depend on `korlibs-datastructure`. And if it just uses a couple of functions, then we can do the same: copy them as internal functionality.

holloszaboakos:
That doesn't solve circular dependency problem (korlibs-datastructure also needs dependency for foundation). However as we already discussed on discord, we can duplicate code and make that internal.

PR #2113, opened 2024-01-03
Extract korlibs-datastructure

holloszaboakos:
Partial work on https://github.com/korlibs/korge/issues/2040


soywiz (changes requested):
Hi @holloszaboakos thanks for the PR, let me comment a few adjustments that should be performed:

## `korlibs.platform` package

I have checked where it is used. It seems that it is used only in a class in the JS target, and in tests:

Screenshot 2024-01-03 at 11 23 18

This is used mainly to detect runtimes etc.

Let's create a `korlibs-platform` module:

It can be included in tests like this:

```
commonTestApi(project(":korlibs-platform"))
```

Let's keep the `korlibs.datastructure.event` in the `korge-foundation` for now. That way the main source set can be kept without depending on `korlibs-platform`. As it is not required.

## Buffer, UShortArrayInt, UByteArrayInt

They are not used, and since it is internal the arraycopy variants using them should be removed. You can search for references and delete the unused ones.

## Only `korlibs.platform`

Please, move the remaining stuff from other packages: InternalMath and InternalMemory (arraycopy etc.) to the `korlibs.platform.internal` package to clearly state this is internal utilities to that package and module.

holloszaboakos:
@soywiz Hi! I am not 100% sure what are you asking. We are going to split the foundation further. Therefor this issue will be completly resolved when platfor fets its own module. Are you sure we should leave classes temporally in foundation? I would trather do the whole foundation split in one PR instead.

soywiz:
> @soywiz Hi! I am not 100% sure what are you asking. We are going to split the foundation further. Therefor this issue will be completly resolved when platform gets its own module. Are you sure we should leave classes temporally in foundation? I would rather do the whole foundation split in one PR instead.

The EventLoop stuff might be moved later to korge itself where it is used keeping its package. Not directly a datastructure. So yeah, let's keep it in the `korge-foundation` for now. I don't think it makes sense to include it in `korlibs-datastructure`.

Reviewing the whole foundation split would be a lot of work and riskier if some changes happen in the meantime. I would prefer doing one module at a time.

Let's focus on one thing at a time, small steps:

* The `korlibs-platform` doesn't have dependencies to anything else. That could be the best first split. datastructures only use it in tests.
* Place only the functions that are used in the `korlibs.datastructure.internal` package instead of `korlibs.mathinternal` etc. it should be a few functions like ilog2, some arraycopy signatures and not too much else.

soywiz:
Please, to avoid having duplicate code that is unused, remove Buffer, Buffer-related arraycopy variants and UByte/UShortArrayInt:

Screenshot 2024-01-03 at 14 30 21

soywiz:
And all the functions that appear grey (unused):

Screenshot 2024-01-03 at 14 32 18

The rest that are still used, should be marked as `internal` to not be exposed.

holloszaboakos:
@soywiz  I am going to make the deletion of not needed features as a last step after everithing build correctly and all tests succeed.

settings.gradle.kts
```diff
@@ -41,3 +46,4 @@ if (System.getenv("DISABLE_SANDBOX") != "true") {
 if (!inCI || System.getenv("ENABLE_BENCHMARKS") == "true") {
     include(":korge-benchmarks")
 }
+include("korlibs-platform")
```
soywiz, line 49: This line is duplicated

korge-core/build.gradle.kts
```diff
@@ -5,6 +5,8 @@ dependencies {
         //add("androidMainApi", "androidx.javascriptengine:javascriptengine:1.0.0-alpha05")
         //add("androidMainApi", "com.google.guava:guava:31.0.1-android")
     }
+    commonMainApi(project(":korlibs-datastructure"))
+    commonMainApi(project(":korlibs-platform"))
```
soywiz, line 9: This should be removed. I mean, the idea is that those artifacts are transitively resolved via `korge-foundation` that is already included.

korlibs-datastructure/src/korlibs/datastructure/_Datastructure_event.kt
```diff
@@ -88,7 +87,7 @@ open class SyncEventLoop(
             if (running) {
                 timedTasks.add(task)
             } else {
-                Console.warn("WARNING: QUEUED TASK time=$time interval=$interval without running")
+                //TODO: add back logger Console.warn("WARNING: QUEUED TASK time=$time interval=$interval without running")
```
soywiz, line 90: This class/package should be kept in the korge-foundation for now, and there the logger package should be available.

settings.gradle.kts
```diff
@@ -13,6 +13,9 @@ pluginManagement {
         maven { url = uri("https://oss.sonatype.org/content/repositories/snapshots/") }
     }
 }
+plugins {
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.5.0"
+}
```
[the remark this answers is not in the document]
holloszaboakos: I did not add it manually. It must have been intelij. Resolved.

korlibs-platform/src@android/korlibs/platform/Current.android.kt
```diff
@@ -1,8 +1,8 @@
 package korlibs.platform
 
 import android.os.*
-import korlibs.korge_foundation.*
-import korlibs.memory.*
+//import korlibs.korge_foundation.*
```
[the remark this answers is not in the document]
holloszaboakos: Unresolved reference: korge_foundation
Unresolved reference: memory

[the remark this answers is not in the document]
holloszaboakos: something is not generating properly

holloszaboakos:
The last thing to do is moving the event stuff back to foundation

holloszaboakos:
@soywiz It is complete

buildSrc/src/main/kotlin/korlibs/korge/gradle/targets/android/AndroidDirect.kt
```diff
@@ -84,9 +82,7 @@ fun Project.configureAndroidDirect(projectType: ProjectType, isKorge: Boolean) {
         }
 
         buildFeatures.apply {
-            if (project.name == "korge-foundation") {
-                buildConfig = true
-            }
+            buildConfig = true
```
soywiz, line 85: Now that we know how it works, can you restore this if with `korlibs-platform`?

korge-foundation/src@jvmAndroid/korlibs/event/_Datastructure_event.jvmAndroid.kt
```diff
@@ -1,6 +1,6 @@
 @file:Suppress("PackageDirectoryMismatch")
 
-package korlibs.datastructure.event
+package korlibs.event
```
soywiz, line 3: This is wrong. It should be `korlibs.datastructure.event` even if it is in the `korlibs-foundation` module. Otherwise we would be breaking compatibility. And we don't want that in this PR

soywiz:
I have noticed that there are a few files that are deleted instead of moved:

image

Then re-added, any chance they could appear here as moved so it is easier to review?

korlibs-datastructure/test/korlibs/datastructure/ByteArrayDequeTest.kt
```diff
@@ -26,7 +25,9 @@ class ByteArrayDequeTest {
         data.write(byteArrayOf(0x99.toByte(), 0x88.toByte(), 0x33, 0x44, 0x55, 0x66, 0xAA.toByte()))
         assertEquals(8, data.availableRead)
         assertEquals(0, data.availableWriteWithoutAllocating)
-        assertEquals(listOf(0x7F, 0x99, 0x88, 0x33, 0x44, 0x55, 0x66, 0xAA), data.readBytesUpTo(8).map { it.toInt() and 0xFF })
+        assertEquals(
+            listOf(0x7F, 0x99, 0x88, 0x33, 0x44, 0x55, 0x66, 0xAA),
+            data.readBytesUpTo(8).map { it.toInt() and 0xFF })
```
soywiz, line 30: Why this change? Can we have moved files without changing imports or reformatting? We can do reformatting / optimizing imports in a separate PR if required.

korlibs-platform/src/korlibs/platform/Current.kt
```diff
@@ -4,7 +4,7 @@ internal expect val currentOs: Os
 internal expect val currentRuntime: Runtime
 internal expect val currentArch: Arch
 internal expect val currentIsDebug: Boolean
-internal expect val currentIsLittleEndian: Boolean
+expect val currentIsLittleEndian: Boolean
```
soywiz, line 7: This shouldn't be used.

korlibs-datastructure/test/korlibs/datastructure/ByteArrayDequeTest.kt
holloszaboakos: I always end the PR by formatting. Turning back now is too much effort. If you prefer other formatting than I suggest edding automated formatting so it happens when someone pushes the commit.

holloszaboakos:
> I have noticed that there are a few files that are deleted instead of moved:
> 
> image
> Then re-added, any chance they could appear here as moved so it is easier to review?

These files were only moved, not changed. They should be fine. The only event test is moved back to foundation.

soywiz:
Superseded by:
* https://github.com/korlibs/korge/pull/2122
* https://github.com/korlibs/korge/pull/2123

Issue #2112, opened 2024-01-02
Why korlibs-crypto depends on korlibs-time?


morki:
I did not find any real dependency in code, is it really needed?

soywiz:
Maybe it was an error. Can you make a PR removing the dependency? If CI passess, it should be fine.

PR #2114, opened 2024-01-03
Remove time from crypto dependencies (#2112)

InsanusMokrassar:
Fixes https://github.com/korlibs/korge/issues/2112


soywiz:
Fixes https://github.com/korlibs/korge/issues/2112

soywiz:
Thanks!

PR #2118, opened 2024-01-03
Fixes packageJvmFatJar with gradle configuration cache

soywiz:
Try to fix https://github.com/korlibs/korge-hello-world/issues/29

Test packageJvmFatJar with configuration-cache on CI on e2e tests


soywiz:
Let's reproduce the issue on CI first

PR #2126, opened 2024-01-12
Implement ASE 9-patch reading in KorGE

soywiz:
@jobe-m 

Screenshot 2024-01-12 at 20 32 06


jobe-m:
Awesome, thanks :)
You're really fast. I will try it out next Monday during the stream.
I hope it does not break other features of aseprite file loading.

soywiz:
> Awesome, thanks :) You're really fast. I will try it out next Monday during the stream. I hope it does not break other features of aseprite file loading.

It shouldn't. Tests are passing, and there is a good battery of tests there. This is an optional nullable ninePatchSlice that is only created when available.

PR #2129, opened 2024-01-13
Allow providing pre-allocated BufferedImage for awtConvertImage

Kietyo:
BufferedImage itself has internal data buffers that it allocates everytime it's allocated, so this helps reduce memory usage and garbage collection if I'm calling this function often.

For context I have a personal project that grabs the image (a BufferedImage) from HDMI capture device, than sends it over to clients over my network so I can view my surveillance monitor in a different area. Each time I grab the image, I need to convert the image using awtConvertImage so that I can use Korge's Bitmap utils; notably QOI encoding to reduce the size of the data before sending it over the network. Currently, each time awtConvertImage is called, it creates a new BufferedImage.


soywiz (approved):
LGTM

PR #2132, opened 2024-01-16
Introduce korlibs-concurrent (Lock + NativeThread) + add logger, datastructure and platform to K/N desktop targets

soywiz:
/CC @holloszaboakos


korlibs-logger/src@js/korlibs/logger/Logger.js.kt
```diff
@@ -20,14 +19,17 @@ actual object DefaultLogOutput : Logger.Output {
     override fun output(logger: Logger, level: Logger.Level, msg: Any?) = Logger.ConsoleLogOutput.output(logger, level, msg)
 }
 
-external private val process: dynamic
-external private val Deno: dynamic
+private external val process: dynamic
+private external val Deno: dynamic
+
+private val isDenoJs: Boolean by lazy { js("(typeof Deno === 'object' && Deno.statSync)").unsafeCast<Boolean>() }
```
holloszaboakos, line 25: Why do you duplicate platform code in logger? Why is it a problem to depend on platform?

korlibs-logger/src@linux/korlibs/logger/Logger.linux.kt
```diff
@@ -0,0 +1,28 @@
...
+    override fun output(logger: Logger, level: Logger.Level, msg: Any?) = Logger.ConsoleLogOutput.output(logger, level, msg)
+}
+
+internal actual val miniEnvironmentVariables: Map<String, String> by lazy {
+    getEnvs()
+}
+
+private fun getEnvs(): Map<String, String> {
+    val out = LinkedHashMap<String, String>()
+    val env = platform.posix.__environ
```
holloszaboakos, line 19: We should avoid so short variable names. It really makes the code hard to read.

korlibs-logger/src@mingw/korlibs/logger/Logger.mingw.kt
```diff
@@ -0,0 +1,42 @@
...
+
+actual object Console : BaseConsole()
+
+actual object DefaultLogOutput : Logger.Output {
+    override fun output(logger: Logger, level: Logger.Level, msg: Any?) = Logger.ConsoleLogOutput.output(logger, level, msg)
+}
+
+internal actual val miniEnvironmentVariables: Map<String, String> by lazy { getEnvs() }
+
+private fun readStringsz(ptr: CPointer<WCHARVar>?): List<String> {
```
holloszaboakos, line 16: Is the z in the function name on purpose?

korlibs-logger/src@js/korlibs/logger/Logger.js.kt
soywiz: Because just for two lines, I don't see adding an extra dependency in this case.

korlibs-logger/src@mingw/korlibs/logger/Logger.mingw.kt
soywiz: Yes. The stringz is typically from C-languags world. It is a string that ends with a '\0' byte. Typically in C strings strings work like that instead of having the size attached, they just end by '\0'

korlibs-datastructure/src/korlibs/datastructure/_Datastructure_lock.kt
```diff
@@ -2,62 +2,18 @@
...
-}
-
-fun BaseLock.wait(time: TimeSpan, precise: Boolean): Boolean {
-    return if (precise) waitPrecise(time) else wait(time)
-}
+import korlibs.concurrent.lock.waitPrecise as waitPreciseConcurrent
+import korlibs.concurrent.lock.wait as waitConcurrent
+
+@Deprecated("Use korlibs.concurrent.lock package")
+typealias BaseLock = korlibs.concurrent.lock.BaseLock
```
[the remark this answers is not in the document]
soywiz: Yes for both questions. The thing is that we want to keep at least source-code compatibility for at least a few versions.

korlibs-concurrent/src/korlibs/concurrent/lock/Lock.kt
```diff
@@ -0,0 +1,64 @@
...
+import kotlin.time.Duration.Companion.seconds
+
+interface BaseLock {
+    fun notify(unit: Unit = Unit)
+    fun wait(time: Duration): Boolean
+    //fun lock()
+    //fun unlock()
+}
+
+//typealias Lock = BaseLock
```
holloszaboakos, line 17: We should not leave in comment code in PR. YAGNI: You aren't gonna need it. If you do, you can readd it from earlier commit.

korlibs-logger/src@linux/korlibs/logger/Logger.linux.kt
soywiz: We could control env local variable, but not __environ since that comes from the operating system

korlibs-concurrent/src/korlibs/concurrent/lock/Lock.kt
```diff
@@ -0,0 +1,64 @@
+@file:Suppress("PackageDirectoryMismatch")
+
+package korlibs.concurrent.lock
+
+import korlibs.concurrent.thread.*
```
holloszaboakos, line 5: * imports are not adviced since they make it hard to determine which classes are from which imports.

soywiz: Rationale behind the star imports:

KorGE uses star imports because it relies a lot on extension members. It was a pain in the ass adding all those members manually for people starting. Users typically don't care where classes come from as long as the code works.

korlibs-concurrent/src/korlibs/concurrent/lock/Lock.kt
soywiz: This is old code that was like that, I'm okay with removing those lines. So feel free to do so.

korlibs-logger/src@js/korlibs/logger/Logger.js.kt
soywiz: Hopefully in Kotlin 2 we will have the ability to differentiate runtimes integrated in the stdlib at some point.

korlibs-concurrent/src/korlibs/concurrent/lock/Lock.kt
holloszaboakos: Yes, but it makes it pain in the ass to refactor. For our current task forexample it is important to see which methods are used from which packages, when we want to reduce the dependency towards a package.

soywiz: Why it is a pain in the ass? Doesn't `@Deprecated` has a parameter to specify replacement to do it automatically inside the IDE?

PR #2144, opened 2024-02-06
Fix putSliceWithBorders extention function for Bitmap class

jobe-m:
The function putSliceWithBorders is used to create an extruded tileset in LdtkView. The tiles were not extruded correctly into top and bottom left corners. This commit fixes that. It also adds a unit test which reproduces the error.

Incorrectly expanded example of a rectangle:

image

### DIFF
--- a/korge-core/src/korlibs/image/bitmap/BitmapExt.kt
+++ b/korge-core/src/korlibs/image/bitmap/BitmapExt.kt
@@ -53,8 +53,8 @@ fun Bitmap.putSliceWithBorder(x: Int, y: Int, bmp: BmpSlice, border: Int = 1) {
     // Vertical replicate
     for (n in 1..border) {
         val rwidth = width + border * 2
-        this.copy(x, y, this, x, y - n, rwidth, 1)
-        this.copy(x, y + height - 1, this, x, y + height - 1 + n, rwidth, 1)
+        this.copy(x - border, y, this, x - border, y - n, rwidth, 1)
+        this.copy(x - border, y + height - 1, this, x - border, y + height - 1 + n, rwidth, 1)
     }
 }
 
--- a/korge-core/test/korlibs/image/bitmap/BitmapSliceTest.kt
+++ b/korge-core/test/korlibs/image/bitmap/BitmapSliceTest.kt
@@ -12,6 +12,7 @@ import korlibs.math.geom.vector.*
 import korlibs.platform.*
 import kotlin.test.Test
 import kotlin.test.assertEquals
+import kotlin.test.assertTrue
 
 class BitmapSliceTest {
     val logger = Logger("BitmapSliceTest")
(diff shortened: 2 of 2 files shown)


korge-core/src/korlibs/image/bitmap/BitmapExt.kt
```diff
@@ -53,8 +53,8 @@ fun Bitmap.putSliceWithBorder(x: Int, y: Int, bmp: BmpSlice, border: Int = 1) {
     // Vertical replicate
     for (n in 1..border) {
         val rwidth = width + border * 2
-        this.copy(x, y, this, x, y - n, rwidth, 1)
-        this.copy(x, y + height - 1, this, x, y + height - 1 + n, rwidth, 1)
+        this.copy(x - border, y, this, x - border, y - n, rwidth, 1)
```
soywiz, line 56: Shouldn't be `rwidth + border * 2`?

jobe-m: What do you exatly mean should I put that term?
I did the fix in a rush. Let me check it carefully with a bit more time on Friday :)
I will create also a "better" (more colorful) tile for visually checking if the expansion of the border works as we expect.

soywiz:
```
fun Bitmap.copy(srcX: Int, srcY: Int, dst: Bitmap, dstX: Int, dstY: Int, width: Int, height: Int)
```

I mean, if border was not taken into account and if position is displaced now, width might need to be extended?

For example, let's imagine a segment: x=10, width=10 (right = 20)
If I want to add one pixel on both sides to the line: x=9, width=12 (right=21)

So subtracting `N` to the `x` (`left`), requires adding `N*2` to the width, or N to the `right` coordinate.
In the case we are extending the same amount of pixels on both sides of the segment.

Not sure if that was the case here, but asked just in case.

korge-core/test/korlibs/image/bitmap/BitmapSliceTest.kt
```diff
@@ -27,6 +27,25 @@ class BitmapSliceTest {
         assertEquals("Rectangle(x=24, y=24, width=24, height=24)", bmp.sliceWithSize(16, 16, 32, 32).sliceWithSize(8, 8, 40, 40).bounds.toString())
     }
 
+    @Test
+    fun testPutSliceWithBorder() {
```
soywiz, line 31: Regarding test. Maybe the Bitmap32 can be converted into a Bitmap1/Bitmap8 and then: https://github.com/korlibs/korge/blob/11993ae7828c96ee261b5d610904c8ca5430b8d2/korge-core/test/korlibs/image/bitmap/Bitmap1Test.kt#L8

https://github.com/korlibs/korge/blob/65ca3e1e284fbb4bffb413a1ba687437befef4bc/korge-core/src/korlibs/image/bitmap/Bitmap8.kt#L49

`bmp.tryToExactBitmap8().toLines(".*#").joinToString("\n")`

or use a reference image. Since we are in korge-core, we should be able to generate screenshots.

korge-core/src/korlibs/image/bitmap/BitmapExt.kt
jobe-m: You are right. I checked now after my Friday stream again and my implementation actually adds the border twice. So, if I add border = 2 than 4 pixels are added to each side of the tile :)

I need to check again...

korge-core/test/korlibs/image/bitmap/BitmapSliceTest.kt
jobe-m: I checked using Bitmap1 in the unit test but `putWithBorder()` is not implemented for Bitmap1. What seems to be missing is the implementation of `slice()` for Bitmap1.

korge-core/src/korlibs/image/bitmap/BitmapExt.kt
jobe-m: I checked again now and I could not find a problem anymore with my initial changes. Here is a screenshot for border = 1:

image

So it expands correctly by one pixel in each direction. I checked that visually with border = 2, 3 and 5 and it was correctly expanding with 2 resp. 3 and 5 pixels.

korge-core/test/korlibs/image/bitmap/BitmapSliceTest.kt
jobe-m: Since `Slice()` is not implemented for Bitmap1 I would like to keep the unit test as it is now.

jobe-m: I can try with Bitmap8 indeed and create an imput and output image with indexes...

jobe-m:
Now unit test uses Bitmap8 and has a better visual test

soywiz (approved):
LGTM

PR #2146, opened 2024-02-09
util, memory and math-core package separation

holloszaboakos:
I tried te separate the memory modul from foundation.
I had to leave some features there because of dependency on the Half class from number.
I also had to separate the util package and create tha math-core modul for broadly used math functionality.
I included memory and math-core in datastructures, so I could remove meny internal functionality.


korlibs-math-core/build.gradle.kts
```diff
@@ -0,0 +1,19 @@
...
+        "https://raw.githubusercontent.com/korlibs/korge/main/korlibs-math-core/LICENSE"
+    )
+}
+
+dependencies {
+    commonMainApi(libs.kotlinx.coroutines.core)
+    commonMainApi(libs.kotlinx.atomicfu)
+    commonTestApi(libs.kotlinx.coroutines.test)
+    commonMainApi(libs.kotlinx.atomicfu)
+    commonTestApi(libs.kotlinx.coroutines.test)
```
soywiz, line 18: Duplicated lines. Why kotlinx.coroutines is required here?

soywiz:
Thanks @holloszaboakos !

For the sake of being easier to review. Let's split this PR: 
* `inject` in one PR
* `math-core` in other PR
  * After `math-core` is merged, let's update `korlibs-memory` in other PR
  * After `math-core` is merged, let's update `korlibs-util` in other PR

holloszaboakos:
I will, but first comment on the direction pls. Are these modules and their name fine by you? What about the memory code left in foundation?

soywiz:
> I will, but first comment on the direction pls. Are these modules and their name fine by you? What about the memory code left in foundation?

Look reasonable to me. Since we are not going to change change packages or class names, it would be just renaming a folder, and changing a string in 2 o 3 files. So we can change our mind easily if we find better names. But as said, I believe they are reasaonable already.

holloszaboakos:
it will be sliced into multiple PR-s

PR #2153, opened 2024-02-16
Remove prints from StandardBasePathsDarwin.kt

juraj-hrivnak:
```
StandardBasePathsDarwin.resourcesFolder:
  ->
```
These prints were probably left here unintentionally.


soywiz:
Thanks!

PR #2154, opened 2024-02-19
Improvements for UITextInput

### LINKED ISSUE #2150: FR: Add a blinking cursor for uiTextInput when user clicks on it
Currently when u click in the input, no blinking cursor appears so it's not clear if it's currently being focused on.

Example when I click the text input for "Login" (no blinking cursor so u can't tell):

image

Kietyo:
- Fixes #2150 
- Added setting to control caret blink duration or to make it static. Also useful for tests.


soywiz:
So we have two issues here:

* No caret appearing when focusing an empty textbox
* Allow to specify whether the caret will block or not

Could you create two separate PRs?

korge/src/korlibs/korge/text/TextEditController.kt
```diff
@@ -24,19 +24,24 @@ import korlibs.time.*
 import kotlin.math.*
 import kotlin.text.isLetterOrDigit
 
+data class TextInputSettings(
+    // If true, render static non-blinking caret rather than blinking caret.
+    val useStaticCaret: Boolean = false
```
soywiz, line 29: I would rename it to blinking caret = true to name it in a positive way. 

Or alternatively, change it to configure the duration of the blinking:

```
val blinkingTime: Duration? = 1.seconds
```

I would go for the duration, that way we can use 0 or null for static non-blinking.
But as stated in the other comment, I believe this is a separate issue not related to the title of the PR "Support caret on empty text input"

Kietyo: I'm willing to change the api but not willing to split into separate PRs as it's a PITA. I can just update the title to "Improve UITextInput" to make it more general.

soywiz: Works for me

Kietyo:
Done.

korge/test/korlibs/korge/ui/UITextInputTest.kt
```diff
@@ -82,7 +82,7 @@ class UITextInputTest : ViewsForTesting() {
 
     inner class TextInputTester(val stage: Stage) {
         val log = arrayListOf<String>()
-        val textInput = stage.uiTextInput()
+        val textInput = stage.uiTextInput() {}
```
soywiz, line 85: What's this change?

korge/src/korlibs/korge/text/TextEditController.kt
```diff
@@ -25,8 +25,9 @@ import kotlin.math.*
 import kotlin.text.isLetterOrDigit
 
 data class TextInputSettings(
-    // If true, render static non-blinking caret rather than blinking caret.
-    val useStaticCaret: Boolean = false
+    // Controls caret blinking duration.
+    // If null, then caret will be static.
+    val caretBlinkingDuration: TimeSpan? = 0.5.seconds,
```
soywiz, line 30:
```kotlin
companion object {
  val STATIC_CARET = TextInputSettings(caretBlinkingDuration = null)
}
```

korge/test/korlibs/korge/ui/UITextInputTest.kt
```diff
@@ -11,7 +11,7 @@ import kotlin.test.*
 class UITextInputTest : ViewsForTesting() {
     @Test
     fun testBackspace() = viewsTest {
-        val textInput = uiTextInput()
+        val textInput = uiTextInput() {}
```
soywiz, line 14: What's this change?

korge/src/korlibs/korge/view/Text.kt
```diff
@@ -230,6 +230,10 @@ open class Text(
         return _textMetricsResult ?: error("Must ensure font is resolved before calling getGlyphMetrics")
     }
 
+    fun getGlyphMetricsForSpaceCharacter(): TextMetricsResult {
+        return font.getOrNull()?.getTextBoundsWithGlyphs(fontSize, " ", renderer, alignment) ?: error("Must ensure font is resolved before calling getGlyphMetrics")
+    }
```
soywiz, line 235: Since we are adding it, what about?

```kotlin
getGlyphMetricsForText(text: String)
```

So it is more generic. Or at least make it internal.

Issue #2164, opened 2024-02-26
graphics{} on uiVerticalStack is not checking bounds


rafi0:
Code like this:
```
uiVerticalStack {
    position(400, 500)
    uiText("a")
    uiText("a")

    graphics {
      stroke(Colors.DARKGREY, info = StrokeInfo(thickness = 1.5f)) {
        moveTo(15, 15)
        lineTo(25, 25)
      }
    }
    uiText("a")
    uiText("a")
  }
```

gives this result:
![Zrzut ekranu 2024-02-26 214009](https://github.com/korlibs/korge/assets/33195725/badfafd1-3784-4072-8549-eea8ff11b400)

and i would rather expect something like this:
![Zrzut ekranu 2024-02-26 214143](https://github.com/korlibs/korge/assets/33195725/86be502b-f1e6-42c2-9dd2-9d044fef52b3)

so im guessing we need UiGraphics for this case?

PR #2166, opened 2024-02-26
Update UIContainerLayouts.kt so graphics{} work better

rafi0:
Fixes https://github.com/korlibs/korge/issues/2164


soywiz:
Thanks! We will need to do the same for the rest, including UIHorizontalStack, etc.
Also we will need to add a test that verifies Y positions flor these views. And maybe a reference tests. Instead of texts, with solid rects that are less likely to fail due slight vector inconsistencies.

soywiz:
Fixes https://github.com/korlibs/korge/issues/2164

soywiz (changes requested):
We need two things:

* Update UIHorizontalStack too.
* Add a test that checks the new behaviour.

My suggestion for the test is the following:

```
@Test
fun testUIVerticalStack() {
  val stack = UIVerticalStack()
  val view1 = stack.solidRect(10, 10, Colors.RED)
  val view2 = stack.graphics { fill(Colors.RED) { rect(10, 10, 20, 20) } }
  val view3 = stack.solidRect(10, 10, Colors.RED)
  stack.relayout()
  assertEquals("...", "${view1.y.niceStr},${view2.y.niceStr},${view3.y.niceStr}")
}
```

By applying/reverting this change the view3.y should be different. So that will cover that we have fixed the issue.

soywiz (approved):
LGTM

soywiz:
Thank you very much!

PR #2170, opened 2024-02-29
Korge memory split

holloszaboakos:


korlibs-memory/src/korlibs/memory/ByteArrayBuilder.kt
```diff
@@ -10,7 +9,7 @@ import kotlin.math.*
 public class ByteArrayBuilder(public var data: ByteArray, size: Int = data.size, public val allowGrow: Boolean = true) {
     public constructor(initialCapacity: Int = 4096) : this(ByteArray(initialCapacity), 0)
 
-    private var _size: Int = size
+    var _size: Int = size
```
soywiz, line 12: This shouldn't be exposed.

korlibs-memory/src/korlibs/memory/ByteArrayBuilder.kt
```diff
@@ -35,11 +34,11 @@ public class ByteArrayBuilder(public var data: ByteArray, size: Int = data.size,
         }
     }
 
-    private fun ensureCount(count: Int) {
+    fun ensureCount(count: Int) {
```
soywiz, line 37: Ditto

soywiz (changes requested):
I'm fine with extracting the `Half` functionality from memory, to not add that extra dependency. But private functionaluty shouldn't be exposed in the process.

As commented `Half.rawBits.toShort` and `Half.fromBits` using with `s16` variants should do the trick.

korlibs-memory/src/korlibs/memory/ByteArrayBuilder.kt
holloszaboakos: This is the problem with the half structure. If I do not expose code like this, I can't make those functions extention functions. This was my problem in the original PR. I need some help with it, because I don't know how the Half stuff works and why.

korlibs-memory/src/korlibs/memory/ByteArrayBuilder.kt
holloszaboakos: ditto

korge-foundation/src/korlibs/memory/ByteArrayBuilderExtentions.kt
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Analogous to [StringBuilder] but for [ByteArray]. Allows to [append] values to end calling [toByteArray].
+ * Provides some methods like [s16LE] or [f32BE] to append specific bit representations easily.
+ */
+fun ByteArrayBuilder.f16(v: Half, little: Boolean): ByteArrayBuilder = this.apply { prepare(2) { data.setF16(_size, v, little) } }
+fun ByteArrayBuilder.f16LE(v: Half): ByteArrayBuilder = this.apply { prepare(2) { data.setF16LE(_size, v) } }
+fun ByteArrayBuilder.f16BE(v: Half): ByteArrayBuilder = this.apply { prepare(2) { data.setF16BE(_size, v) } }
+
+fun ByteArrayBuilderLE.f16(v: Half): ByteArrayBuilder = bab.f16LE(v)
+fun ByteArrayBuilderBE.f16(v: Half): ByteArrayBuilder = bab.f16BE(v)
```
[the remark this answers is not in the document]
holloszaboakos: I see! This solves it, thanks!

korge-foundation/src/korlibs/memory/ByteArrayReaderExtention.kt
```diff
@@ -0,0 +1,10 @@
+package korlibs.memory
+
+import korlibs.number.Half
+
+
+fun ByteArrayReader.f16(little: Boolean): Half = move(2) { getF16(it, little) }
+fun ByteArrayReader.f16LE(): Half = move(2) { getF16LE(it) }
+fun ByteArrayReader.f16BE(): Half = move(2) { getF16BE(it) }
+fun ByteArrayReaderLE.f16(): Half = bar.f16LE()
+fun ByteArrayReaderBE.f16(): Half = bar.f16BE()
```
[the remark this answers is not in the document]
holloszaboakos: Nice!

korlibs-memory/build.gradle.kts
```diff
@@ -0,0 +1,21 @@
...
+
+dependencies {
+    commonMainApi(libs.kotlinx.coroutines.core)
+    commonMainApi(libs.kotlinx.atomicfu)
+    commonTestApi(libs.kotlinx.coroutines.test)
+    commonMainApi(project(":korlibs-math-core"))
+    commonMainApi(project(":korlibs-platform"))
+    commonMainApi(project(":korlibs-util"))
+    commonMainApi(project(":korlibs-crypto"))
+}
```
soywiz, line 21: We should evaluate those.

* kotlinx-coroutines are not used at all
* math-core might be used
* util and crypto probably not?

korge-foundation/src/korlibs/memory/ByteArrayExtention.kt
```diff
@@ -0,0 +1,17 @@
+package korlibs.memory
+
+import korlibs.number.Half
+
+private fun ByteArray.u8(offset: Int): Int = this[offset].toInt() and 0xFF
+private inline fun ByteArray.get16LE(offset: Int): Int = (u8(offset + 0) shl 0) or (u8(offset + 1) shl 8)
+private inline fun ByteArray.get16BE(offset: Int): Int = (u8(offset + 1) shl 0) or (u8(offset + 0) shl 8)
```
soywiz, line 7: Aren't these exposed publicly already in other package?

korlibs-memory/build.gradle.kts
holloszaboakos: util and math-core are used. We should avoid copiing stuff to private. I check the other two

korge-foundation/src/korlibs/memory/ByteArrayExtention.kt
holloszaboakos: I dont know about that. They were private in the original class

korlibs-memory/build.gradle.kts
soywiz: Okay, let's keep it as it is for now. We can optimize dependencies later.

soywiz:
Thanks!

PR #2225, opened 2024-04-30
Fixes runJs, deduplicates template and adds <noscript> to index.html template

soywiz:


soywiz:
ES2015 output reduces `:korge-sandbox:runJs` from 2.5MB to 1.9MB in release mode (esbuild minification without extra closure compiling)

PR #2319, opened 2025-05-02
removeChildAt in O(n) instead of O(n^2) and simplify removeChild

reosfire:
Current code working in O(n *  c) where `n` is children count and `c` is count we are going to remove.
This is the optimization to O(n)

Also I don't understand how is it possible that child view from __children list has other container as a parent. Is it expected behavior? If not then maybe the code should throw instead of returning everywhere?

### DIFF
--- a/korge/src/korlibs/korge/view/Container.kt
+++ b/korge/src/korlibs/korge/view/Container.kt
@@ -180,10 +180,30 @@ open class Container(
     fun removeChildAt(index: Int): Boolean =
         removeChild(getChildAtOrNull(index))
 
-    // @TODO: Optimize
-    fun removeChildAt(index: Int, count: Int) {
-        repeat(count) { removeChildAt(index) }
+    fun removeChildAt(index: Int, count: Int): Int {
+        var actuallyRemoved = 0
+        while (actuallyRemoved < count) {
+            val view = getChildAtOrNull(index + actuallyRemoved)
+            if (view?.parent !== this) break
+            actuallyRemoved++
+        }
+
+        for (removedIndex in index until index + actuallyRemoved) {
+            getChildAtOrNull(removedIndex)?.let { view ->
+                view.parent = null
+                view.index = -1
+            }
+        }
+
+        __children.removeRange(index, index + actuallyRemoved)
+
+        for (i in index until numChildren) __children[i].index -= actuallyRemoved
+
         invalidateZIndexChildren()
+        invalidateContainer()
+        invalidateLocalBounds()
+
+        return actuallyRemoved
     }
 
     // @TODO: Optimize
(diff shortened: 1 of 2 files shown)


korge/src/korlibs/korge/view/Container.kt
```diff
@@ -180,10 +180,30 @@ open class Container(
...
         removeChild(getChildAtOrNull(index))
 
-    // @TODO: Optimize
-    fun removeChildAt(index: Int, count: Int) {
-        repeat(count) { removeChildAt(index) }
+    fun removeChildAt(index: Int, count: Int): Int {
+        var actuallyRemoved = 0
+        while (actuallyRemoved < count) {
+            val view = getChildAtOrNull(index + actuallyRemoved)
+            if (view?.parent !== this) break
```
soywiz, line 187: Maybe this could be an assertion? This shouldn't happen.

korge/src/korlibs/korge/view/Container.kt
```diff
@@ -177,33 +177,46 @@ open class Container(
...
+        __children.removeAt(index)
+
+        for (i in index until numChildren) __children[i].index = i
+
+        invalidateZIndexChildren()
+        invalidateContainer()
+        invalidateLocalBounds()
+
+        return true
+    }
```
soywiz, line 197:
```kotlin
fun removeChildAt(index:Int) = removeChildAt(index, 1)
```

korge/src/korlibs/korge/view/Container.kt
```diff
@@ -467,16 +483,11 @@ open class Container(
...
-        for (i in view.index + 1 until numChildren) __children[i].index--
-        __children.removeAt(view.index)
-        view.parent = null
-        view.index = -1
-        invalidateZIndexChildren()
-        invalidateContainer()
-        invalidateLocalBounds()
-        return true
+        require(view != null) { "Removed view should not be null" }
+        require(view.parent === this) { "View $view is not a child of this container" }
```
soywiz, line 487: According to the function doc, this should do nothing instead of failing

reosfire: So what do you think is better to fix? Doc or behavior

korge/src/korlibs/korge/view/Container.kt
reosfire: Yeah, that's true. But I think it is better to avoid overhead related with for loop and minimum calculation in the single argument version

korge/src/korlibs/korge/view/Container.kt
soywiz: I would keep previous behaviour

korge/src/korlibs/korge/view/Container.kt
soywiz: Why do you think the overhead in this case relatively to the rest of the operations of the function would be noticeable?

reosfire: Well it probably wouldn't. Ok I'll use your approach

soywiz:
Thanks @reosfire for the PR!

PR #2320, opened 2025-05-06
Reduce buffer allocations

reosfire:
We can just copy data from buffer to buffer instead of making clone when buffers size match.


soywiz:
Thanks!

PR #2326, opened 2025-05-13
Add upload() and wrap methods

reosfire:
It seems pretty reasonable to wrap existing buffer because it provides very convinient methods to manipulate with bytes. Also because the mem property is public buffer content already can be changed from outside but it's impossible to mark this object dirty so I've added upload() method to make it possible

### DIFF
--- a/korge-core/src/korlibs/graphics/AGObjects.kt
+++ b/korge-core/src/korlibs/graphics/AGObjects.kt
@@ -66,6 +66,24 @@ class AGBuffer : AGObject() {
     fun upload(data: Buffer): AGBuffer =
         upload(data, needClone = true)
 
+    /**
+     * Marks this buffer as dirty, so it will be uploaded to the GPU later.
+     */
+    fun upload(): AGBuffer {
+        markAsDirty()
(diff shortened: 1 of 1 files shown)


soywiz (commented):
Functionality looks fine to me. I have a few thoughts on naming:

For example, `upload` could be renamed to `reupload` or `uploadCurrent`? Or just make `markAsDirty` public?

The wrap method is also calling `markAsDirty`. 
Could be named as `upload` as well or `wrap AndUpload` or `setBuffer`?

What do you think?

reosfire:
Pretty hard question I must admit XD. I can't choose between them. Each makes sense to me. AI says to make markAsDirty public and save wrap as is :)

soywiz:
Well, not anything super critical. Ideally names should be as short as possible, but also describing side-effects. Maybe even existing naming with `upload` could be improved.

wrap + public markAsDirty would work for me. But not sure if possible since markAsDirty comes from an upper object. So let's use your upload as it is for now. It should be possible to deprecate and rename later if required.

PR #2328, opened 2025-05-14
Configurable atlas textures packing

reosfire:
Autogenerated atlas texture packing in this pr can be configured through .atlas file in the following manner:
```
#enable-rotation=false
#enable-trimming=false
#padding=1
folder1
folder2
```


buildSrc/src/main/kotlin/korlibs/korge/gradle/processor/KorgeTexturePacker.kt
```diff
@@ -7,6 +7,7 @@ import kotlin.system.*
 
 open class KorgeTexturePacker : KorgeResourceProcessor {
     override fun processFolder(context: KorgeResourceProcessorContext) {
+        context.resourceFolders
```
soywiz, line 10: I guess this line should be removed?

PR #2331, opened 2025-05-20
Add and use vertexAttribIPointer

reosfire:


korge-core/src@js/korlibs/kgl/KmlGlJsCanvas.kt
```diff
@@ -237,6 +239,13 @@ class KmlGlJsCanvas(val canvas: HTMLCanvasElement, val glOpts: dynamic) : KmlGl(
     override fun vertexAttrib4f(index: Int, x: Float, y: Float, z: Float, w: Float): Unit = gl.vertexAttrib4f(index, x, y, z, w)
     override fun vertexAttrib4fv(index: Int, v: Buffer): Unit = gl.vertexAttrib4fv(index, v)
     override fun vertexAttribPointer(index: Int, size: Int, type: Int, normalized: Boolean, stride: Int, pointer: Long): Unit = gl.vertexAttribPointer(index, size, type, normalized, stride, pointer.toInt())
+    override fun vertexAttribIPointer(index: Int, size: Int, type: Int, stride: Int, pointer: Long): Unit {
+        if (webglVersion >= 2) {
+            gl.asDynamic().vertexAttribIPointer(index, size, type, stride, pointer.toInt())
+        } else {
+            TODO()
```
ygdrasil-io, line 246: no fallback on this case ?

ygdrasil-io:
Can you provide more context about what brings the improvement and how did you do the non-regression test?

reosfire:
@ygdrasil-io
I faced a problem when created an Attribute of SInt type and then read it in the shader.
image
image
The problem was that the index took weird values(not values that I've passed to the buffers)

In opengl there are two functions to describe vertex attributes
image
The difference between them is described here
image
When I changed code to what is present in this pr passed int become correct.

In the engine I can't see any places that use Ints as vertex attributes. So there should be very little regressions in client code (only if somebody used Ints in his own shaders and used them as described in docs)

Maybe the better approach would be to add a parameter to the Attribute class which will say which one of this two methods to use

PR #2362, opened 2026-04-28
Update Maven Central coordinates to "org.korge" and switch publishing to vanniktech plugin

jobe-m:


README.md
```diff
@@ -3,7 +3,7 @@
 <!-- BADGES -->
 <p align="center">
 	<a href="https://github.com/korlibs/korge/actions"><img alt="Build Status" src="https://github.com/korlibs/korge/workflows/TEST/badge.svg" /></a>
-    <a href="https://search.maven.org/artifact/com.soywiz.korlibs.korge.plugins/korge-gradle-plugin"><img alt="Maven Central" src="https://img.shields.io/maven-central/v/com.soywiz.korlibs.korge.plugins/korge-gradle-plugin"></a>
+    <a href="https://search.maven.org/artifact/org.korge.gradle.plugins/korge-gradle-plugin"><img alt="Maven Central" src="https://img.shields.io/maven-central/v/org.korge.gradle.plugins/korge-gradle-plugin"></a>
```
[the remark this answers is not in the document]
jobe-m: fixed

buildSrc/src/main/kotlin/korlibs/modules/Signing.kt
```diff
@@ -65,3 +64,23 @@ open class CachedInMemoryPgpSignatoryProvider(signingKey: String?, signingPasswo
...
+    val signingKeyRingFile: String? = System.getenv("ORG_GRADLE_PROJECT_signingSecretKeyRingFile")
+        ?: findProperty("signing.secretKeyRingFile")?.toString()
+    val hasSigningKeys = signingKey != null || signingKeyRingFile != null
+
+    afterEvaluate {
+        tasks.withType(org.gradle.plugins.signing.Sign::class.java).configureEach { signTask ->
+            signTask.onlyIf { hasSigningKeys }
+        }
+    }
+}
```
malliaridis: AFAIK vanniktech plugin does not require signing keys, and uses them only if they are provided. Publishing to `mavenLocal` should therefore not be an issue, even if `maven-publish` is used indirectly.

If I had to challenge this change, I would say that we may not need any of these functions (`configureSigning` and `makeSigningOptional`) as it is handled by vanniktech's plugin already. 

Could you elaborate why this configuration step is required? Does something fail if we do not execute it and don't provide signing keys?

buildSrc/src/main/kotlin/korlibs/root/RootKorlibsPlugin.kt
```diff
@@ -238,7 +249,7 @@ object RootKorlibsPlugin {
                 }
 
                 if (mustPublish) {
-                    plugins.apply("maven-publish")
+                    // maven-publish is applied by vanniktech inside configurePublishing()
                 }
```
malliaridis:
```suggestion
```
I'd remove this block completely then, to avoid having dead code in the project. Git history / blame will show this up anyway if needed. :)

korge-gradle-plugin/src/main/kotlin/korlibs/modules/Signing.kt
```diff
@@ -65,3 +65,23 @@ open class CachedInMemoryPgpSignatoryProvider(signingKey: String?, signingPasswo
...
+        ?: findProperty("signing.secretKeyRingFile")?.toString()
+    val hasSigningKeys = signingKey != null || signingKeyRingFile != null
+
+    afterEvaluate {
+        tasks.withType(org.gradle.plugins.signing.Sign::class.java).configureEach { signTask ->
+            signTask.onlyIf { hasSigningKeys }
+        }
+    }
+}
+
```
malliaridis: Is it posible to remove this duplication?

korge-gradle-plugin/src/main/kotlin/korlibs/root/RootKorlibsPlugin.kt
```diff
@@ -239,7 +250,7 @@ object RootKorlibsPlugin {
                 }
 
                 if (mustPublish) {
-                    plugins.apply("maven-publish")
+                    // maven-publish is applied by vanniktech inside configurePublishing()
                 }
```
malliaridis:
```suggestion
```

Not needed at all, right?

malliaridis (commented):
I went over the changes and it looks quite a lot of work for renaming and configuring a new plugin for publishing. I think there is much to improve here over time.

Once this is done I would recommend moving the gradle plugins into separate repo, as they are not mandatory for KorGE nor do they have anything in common. If they are providing useful stuff for the korge repo or consumers and have to be published for this reason, they should in my opinion follow their own release cycle. This would also decouple korge from gradle plugins and simplify a lot the overall gradle config stuff. Additionally:

- [x] I'd change the URL in the license from `master` branch to `main`, as there is no master branch in the project and it is just an alias for main

I haven't tried it out yet, nor have I searched the project for other references of the old namespace (plan to do that). But what will be very important is

- [x] Add to the README the namespace change, as this is a breaking change and most consumers will have to update their imports.

I would even recommend creating a major release because of that, but if it is well documented, perhaps we can also allow a minor release.

jobe-m:
> I went over the changes and it looks quite a lot of work for renaming and configuring a new plugin for publishing. I think there is much to improve here over time.
> 
> Once this is done I would recommend moving the gradle plugins into separate repo, as they are not mandatory for KorGE nor do they have anything in common. If they are providing useful stuff for the korge repo or consumers and have to be published for this reason, they should in my opinion follow their own release cycle. This would also decouple korge from gradle plugins and simplify a lot the overall gradle config stuff. Additionally:

Yes, that is an good idea. I would do this in separate PRs later.

> I would even recommend creating a major release because of that, but if it is well documented, perhaps we can also allow a minor release.

Fully agree. Let's work on main branch towards release 7.0.0

buildSrc/src/main/kotlin/korlibs/root/RootKorlibsPlugin.kt
jobe-m: Removed

korge-gradle-plugin/src/main/kotlin/korlibs/root/RootKorlibsPlugin.kt
jobe-m: Removed

buildSrc/src/main/kotlin/korlibs/modules/Signing.kt
jobe-m: I removed the code and will check publishing snapshot via CI. Locally artefact upload did not work because of authentication failure.

korge-gradle-plugin/src/main/kotlin/korlibs/modules/Signing.kt
jobe-m: Removed

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
+name: Publish to Maven Central
```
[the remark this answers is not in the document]
jobe-m: DEPLOY.yml is going to be deleted

korge/README.md
```diff
@@ -4,7 +4,7 @@
 <!-- BADGES -->
 <p align="center">
 	<a href="https://github.com/korlibs/korge/actions"><img alt="Build Status" src="https://github.com/korlibs/korge/workflows/CI/badge.svg" /></a>
-    <a href="https://search.maven.org/artifact/korlibs.korge.plugins/korge-gradle-plugin"><img alt="Maven Central" src="https://img.shields.io/maven-central/v/korlibs.korge.plugins/korge-gradle-plugin"></a>
+    <a href="https://search.maven.org/artifact/gradleplugins/korge-gradle-plugin"><img alt="Maven Central" src="https://img.shields.io/maven-central/v/gradleplugins/korge-gradle-plugin"></a>
```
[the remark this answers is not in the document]
jobe-m: fixed

build.gradle.kts
```diff
@@ -11,7 +11,7 @@ RootKorlibsPlugin.doInit(rootProject)
 tasks {
     val checkModulePublication by creating {
         doLast {
-            val publishedKorgeModule = File("${System.getProperty("user.home")}/.m2/repository/com/soywiz/korge/korge/999.0.0.999/korge-999.0.0.999.module")
+            val publishedKorgeModule = File("${System.getProperty("user.home")}/.m2/repository/org/korge/engine/korge/6.1.0-SNAPSHOT/korge-6.1.0-SNAPSHOT.module")
```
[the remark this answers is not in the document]
jobe-m: Unit test related, needs to be updated anyway

buildSrc/src/main/kotlin/korlibs/modules/Publishing.kt
```diff
@@ -1,135 +1,79 @@
...
+            ?: rootProject.findProperty("SONATYPE_USERNAME")?.toString()
+            ?: rootProject.findProperty("sonatypeUsername")?.toString()
+        val pass = System.getenv("SONATYPE_PASSWORD")
+            ?: rootProject.findProperty("SONATYPE_PASSWORD")?.toString()
+            ?: rootProject.findProperty("sonatypePassword")?.toString()
+        if (user != null) project.extensions.extraProperties["mavenCentralUsername"] = user
+        if (pass != null) project.extensions.extraProperties["mavenCentralPassword"] = pass
 
-    //val emptyJar = tasks.createThis<Jar>("emptyJar") {}
+        publishToMavenCentral()
```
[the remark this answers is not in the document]
jobe-m: signing is done by vanniktech plugin

jobe-m:
Checked with publishing `SNAPSHOT`-release to MavenLocal, building and testing my korge-fleks based game application. Looks fine.

PR #2364, opened 2026-05-10
chore: Upgrade project to Gradle and AGP 9 and korlibs 7.0.0-SNAPSHOT

malliaridis:
## Description

As we move forward and plan to upgrade our libraries and tools, major releases require larger refactorings. This PR addresses the major upgrade of Gradle to 9.4.1 and Android Gradle Plugin to 9.1.1, as well as korlibs to 7.0.0-SNAPSHOT.

## Solution

The current AGP upgrade requires a split of the android application plugin and therefore the android entrypoint. This forces us to split the sandbox module into two modules, one is the shared sandbox stuff, and the other the `androidApp` entrypoint module.

The PR removes the usage of the complex setup in `buildSrc` and follows a more gradle-native approach. The existing plugins `korge-gradle-plugin`, `korge-gradle-plugin-common` and `korge-gradle-plugin-settings` are moved to a sub-directory and included as [composite builds](https://docs.gradle.org/current/userguide/composite_builds.html).

The `archive` directory is completely removed, as it is not a meaningful part of the project (it is still available in the git history).

The workflows are flattened to clearly distinguish between each os-specific workflow and their tasks. I chose here duplication for readability and simplicity. Multiple tasks are replaced with just `gradlew check`.

I have compared `buildSrc` with the current `korge-gradle-plugin`. There are minimal differences, and with the move to the composite build, most changes are now identical, with `korge-gradle-plugin` containing more stuff (probably more up-to-date). Therefore, removing `buildSrc` seems like removing a whole duplicated module. You can generate a diff with the IDE and see the actual differences listed to confirm this.

## Open Tasks

- [ ] Comment the removals and what they were doing, how it is done / configured now and where to find the new configs
- [ ] Add notes for potential differences
- [ ] Update github workflows to work with new project configurations
- [ ] Fix / Include e2e modules in builds

## Additional Information

**Important Note when trying this PR**: In order to build and test this PR you have to use Android Studio or Gradle from the CLI. None of the current IntelliJ versions support this build (AGP 9 to be precise) and you have to disable the Android plugin in IntelliJ if you want the Gradle sync to succeed (not tested yet in this project). The AGP versions is not supported, and the proposal of downgrading to 9.0.0-alphaX is not solving the sync issue either. Support for AGP 9 was planned in Q1 of 2026, but we still waiting for it.

**Improtant Note 2**: To run the checks successfully, you have to use a local publication of korlibs with the changes from https://github.com/korlibs/korlibs/pull/201.

Note that the increase of lines of code comes from the API files. In reality the lines of code are reduced by a lot.


malliaridis:
I removed ABI validation from this PR to enable it in a separate PR. ABI validation was not enabled before, and it adds a couple thousand lines of code, which hid the actual impact of this PR.

jobe-m:
Tests are failing with:

```
> Could not resolve all dependencies for configuration ':korge-core:jsNpmAggregated'.
         > Could not find org.korge.korlibs:korlibs-io-stream:7.0.0-SNAPSHOT.
           Searched in the following locations:
             - file:/home/runner/.m2/repository/org/korge/korlibs/korlibs-io-stream/7.0.0-SNAPSHOT/maven-metadata.xml
             - file:/home/runner/.m2/repository/org/korge/korlibs/korlibs-io-stream/7.0.0-SNAPSHOT/korlibs-io-stream-7.0.0-SNAPSHOT.pom
             - https://repo.maven.apache.org/maven2/org/korge/korlibs/korlibs-io-stream/7.0.0-SNAPSHOT/maven-metadata.xml
             - https://repo.maven.apache.org/maven2/org/korge/korlibs/korlibs-io-stream/7.0.0-SNAPSHOT/korlibs-io-stream-7.0.0-SNAPSHOT.pom
             - https://dl.google.com/dl/android/maven2/org/korge/korlibs/korlibs-io-stream/7.0.0-SNAPSHOT/maven-metadata.xml
             - https://dl.google.com/dl/android/maven2/org/korge/korlibs/korlibs-io-stream/7.0.0-SNAPSHOT/korlibs-io-stream-7.0.0-SNAPSHOT.pom
           Required by:
               project ':korge-core'
```

We could publish `7.0.0-SNAPSHOT` to Maven Central (snapshot) from the feature branch in Korlibs (https://github.com/korlibs/korlibs/pull/201) and add here also the maven-snapshot URI to Gradle, right?

malliaridis:
> We could publish 7.0.0-SNAPSHOT to Maven Central (snapshot) from the feature branch in Korlibs (https://github.com/korlibs/korlibs/pull/201) and add here also the maven-snapshot URI to Gradle, right?

Yes, a snapshot release would allow us to run the build for this one. We probably have to add the snapshots repository to the dependency management repositories, but that should be simple enough.

karma.config.d/extra.karma.conf.js
```diff
@@ -1,31 +0,0 @@
...
-      ChromeHeadlessWebGPU: {
-        base: 'ChromeHeadless',
-        flags: [
-          '--enable-unsafe-webgpu', // Enable WebGPU
-          '--enable-features=Vulkan', // Enable Vulkan
-          '--no-sandbox', // Optional: Helps to run Chrome in certain CI environments
-          '--disable-dev-shm-usage', // Optional: Helps to avoid issues with /dev/shm partition in certain CI environments
-          '--headless', // Ensures that Chrome runs in headless mode
-          //'--disable-gpu', // Optional: Disable hardware GPU acceleration (useful in some CI environments)
-          '--remote-debugging-port=9222' // Optional: Allows remote debugging
```
malliaridis, line 14: I have copied the karma configuration from the korlibs repo and moved it to each module that needs one, which is a minimal configuration.

However, these flags are used for development / tests, and could proove useful in the future. I would reenable them afterwards if we noticed issues with the workflows.

korge-benchmarks/build.gradle.kts
```diff
@@ -1,24 +1,28 @@
...
+}
+
+kotlin {
+    applyDefaultHierarchyTemplate()
+    jvm()
+
+    sourceSets {
+        jvmMain.dependencies {
+            api(projects.korge)
+            api(libs.kotlinx.benchmark.runtime)
```
malliaridis, line 14: Many of our dependencies are added as API dependencies, which are passed down transitively and potentially leak unwanted (re)sources. I would address and discuss this in separate PRs and keep the dependencies for now the same.

gradle/libs.versions.toml
```diff
@@ -9,55 +12,77 @@ proguard-gradle = "7.9.1"
...
 dokka = "2.2.0"
-kover = "0.6.1"
+kover = "0.9.8"
 kover-agent = "1.0.776"
 node = "20.12.1"
-android-build-gradle = "8.2.2"
+android-gradle-plugin = "9.2.0"
 gson = "2.14.0"
 gradle-publish-plugin = "2.1.1"
 vanniktech-publish = "0.35.0"
```
jobe-m: We can upgrade vanniktech plugin to `0.36.0` which is the first version which needs Gradle 9.x

jobe-m:
@malliaridis I realized that then running `./gradlew publishToMavenLocal` only the "engine" part of Korge is published. The projects under korge-gradle-plugins are not published. Could it be that the publish part is missing in their build.gradle.kts files?

malliaridis:
> @malliaridis I realized that then running ./gradlew publishToMavenLocal only the "engine" part of Korge is published. The projects under korge-gradle-plugins are not published. Could it be that the publish part is missing in their build.gradle.kts files?

You are right, it is not included because we are using `compositeBuild` (included builds) for the plugins. This means they are not direct sub-projects of the project, and therefore not included. This allows us to prepare the plugins as plugins before the project is initialized, and then use them as plugins in the same project (e.g. in sandbox).

For the publication of the plugins you can and should instead use the module-specific task:

```bash
./gradlew :korge-gradle-plugins:korge-gradle-plugin:publishToMavenLocal
```

The current state / builds of the plugins do not have a version yet (they use `unspecified`), but I will add one before finalizing this PR.

And I believe `korge-core` did not have a publication, right? If not, let me know and I will add that back.

malliaridis:
Should we use a separate release cycle for the plugins and therefore use a different version variable than korge itself from now on, or should it be shared (at least this PR)?

gradle/libs.versions.toml
malliaridis: Done. Thanks for the note. :)

jobe-m:
> For the publication of the plugins you can and should instead use the module-specific task:
> 
> ```shell
> ./gradlew :korge-gradle-plugins:korge-gradle-plugin:publishToMavenLocal
> ```

Thanks, perfect! This is all I need :)

> Should we use a separate release cycle for the plugins and therefore use a different version variable than korge itself from now on, or should it be shared (at least this PR)?

Releasing them separately is OK. I would keep the version in sync for now as long as both come out of the same repo. Later we can decouple the version. But I would suggest that both have the same major version, to keep them aligned.

jobe-m:
> And I believe `korge-core` did not have a publication, right? If not, let me know and I will add that back.

Yes, we need to release `korge-core` also. At least it was published in `6.0.0`.

korge-gradle-plugins/korge-gradle-plugin-common/build.gradle.kts
```diff
@@ -0,0 +1,47 @@
+plugins {
+    id("java")
+    id("java-gradle-plugin")
+    `kotlin-dsl`
+    alias(libs.plugins.vanniktech.mavenPublish)
+}
+
+description = "Multiplatform Game Engine written in Kotlin"
+group = "org.korge.gradle"
```
jobe-m: I used in `6.1.0-SNAPSHOT` release `group = "org.korge.gradleplugins"` in this place. Should we change it or have you a reason to put the code under ".gradle"? Maybe it is common?

korge-gradle-plugins/korge-gradle-plugin-common/build.gradle.kts
```diff
@@ -0,0 +1,47 @@
...
+
```
jobe-m, line 12: Here also version info is needed: `version = libs.versions.korge.get()`

korge-gradle-plugins/korge-gradle-plugin-settings/build.gradle.kts
```diff
@@ -1,31 +1,25 @@
-import korlibs.korge.gradle.targets.android.*
-import korlibs.root.*
-
 description = "Multiplatform Game Engine written in Kotlin"
-group = RootKorlibsPlugin.KORGE_GRADLE_PLUGIN_GROUP
+group = "org.korge.gradleplugins"
```
jobe-m: Here we have `org.korge.gradleplugins`.

korge-gradle-plugins/korge-gradle-plugin-settings/build.gradle.kts
jobe-m, line 39: Version info missing: `version = libs.versions.korge.get()`

korge-gradle-plugins/korge-gradle-plugin/build.gradle.kts
```diff
@@ -0,0 +1,87 @@
+plugins {
+    id("java")
+    id("java-gradle-plugin")
+    `kotlin-dsl`
+    alias(libs.plugins.vanniktech.mavenPublish)
+    alias(libs.plugins.buildconfig)
+}
+
+description = "Multiplatform Game Engine written in Kotlin"
+group = "org.korge.gradle"
```
jobe-m: Here also `org.korge.gradleplugins`?

korge-gradle-plugins/korge-gradle-plugin/src/main/kotlin/korlibs/korge/gradle/targets/android/Android.kt
jobe-m, line 155: Shouldn't we change here both places to use Java 25?

korge-gradle-plugins/korge-gradle-plugin/src/main/kotlin/korlibs/korge/gradle/targets/android/AndroidDirect.kt
```diff
@@ -25,77 +30,51 @@ fun Project.configureAndroidDirect(projectType: ProjectType, isKorge: Boolean) {
...
-        add("androidTestImplementation", "org.jetbrains.kotlin:kotlin-test")
-
-        add("androidTestImplementation", "org.jetbrains.kotlin:kotlin-test")
-        add("androidTestImplementation", "androidx.test:core:1.7.0")
-        add("androidTestImplementation", "androidx.test.ext:junit:1.3.0")
-        add("androidTestImplementation", "androidx.test.espresso:espresso-core:3.7.0")
-        //androidTestImplementation 'com.android.support.test:runner:1.0.2'
-    }
+private fun Project.configureAndroidExecutable(isKorge: Boolean) {
+    project.plugins.apply("com.android.application")
```
jobe-m: It looks like we cannot use this plugin with the new AGP version. Remove line?

korge-gradle-plugins/korge-gradle-plugin/src/main/kotlin/org/korge/kproject/KProjectPlugin.kt
jobe-m: LibraryExtention is deprecated and needs to be exchanged with something else. But I could not figure it out and I could not get it working to use this in a downstream project wiht kproject (e.g. in `korlibs/korge-hello-world` example).

korge-ipc/build.gradle.kts
```diff
@@ -1,31 +1,18 @@
...
-        apiVersion.set(org.jetbrains.kotlin.gradle.dsl.KotlinVersion.fromVersion("2.1"))
-        languageVersion.set(org.jetbrains.kotlin.gradle.dsl.KotlinVersion.fromVersion("2.1"))
-        suppressWarnings.set(true)
-    }
-}
+group = "org.korge.engine"
+version = rootProject.libs.versions.korge.get()
 
 mavenPublishing {
     configure(JavaLibrary(javadocJar = JavadocJar.Empty(), sourcesJar = true))
```
jobe-m, line 18: It looks like we need to change `sourcesJar = true` so something else?

gradle/libs.versions.toml
```diff
@@ -9,55 +12,77 @@ proguard-gradle = "7.9.1"
...
+kotlin-dsl = "6.6.4"
 #kotlinx-coroutines = "1.8.1"
-kotlinx-coroutines = "1.10.2"
+kotlinx-coroutines = "1.11.0"
 kotlinx-serialization = "1.11.0"
 kotlinx-atomicfu = "0.32.1"
-
-kotlinx-benchmark = "0.4.7"
+javaSourceCompatibility = "21"
+javaTargetCompatibility = "21"
```
jobe-m, line 22: Change java version to `25`?

gradle/libs.versions.toml
```diff
@@ -9,55 +12,77 @@ proguard-gradle = "7.9.1"
...
 gson = "2.14.0"
 gradle-publish-plugin = "2.1.1"
-vanniktech-publish = "0.35.0"
+vanniktech-publish = "0.36.0"
 # https://repo.maven.apache.org/maven2/org/jetbrains/intellij/deps/intellij-coverage-agent/
 
+[plugins]
+buildconfig = { id = "com.github.gmazzo.buildconfig", version.ref = "buildconfig" }
+android-multiplatform-library = { id = "com.android.kotlin.multiplatform.library", version.ref = "android-gradle-plugin" }
+android-application = { id = "com.android.application", version.ref = "android-gradle-plugin" }
```
jobe-m, line 37: android-application is not used or deprecated, so I guess we can remove it?

jobe-m:
I put all my findings in comments to not loose them and maybe contribute. They are from trying to get my game project working with `7.0.0-SNAPSHOT`. But I cancelled it now due to lack of knowledge how to proceed and time.

gradle/libs.versions.toml
malliaridis: `android-application` is used by `:korge-sandbox:androidApp`.  This plugin is not compatible with the rest of the multiplatform plugins, and causes all the trouble of splitting existing gradle modules into two separate modules.

gradle/libs.versions.toml
malliaridis: 25 may be the next LTS, but I have seen some lack of support. So it would probably be too early to go with Java 25.

What would should consider is replacing these two values with toolchain configuration, rather than setting source and target compatibility values (old way setting them).

korge-ipc/build.gradle.kts
malliaridis: I haven't figured out what this line configures yet. I will have to look it up to better understand it. I only took it over from previous configuration. I'll share my findings here.

korge-gradle-plugins/korge-gradle-plugin/src/main/kotlin/korlibs/korge/gradle/targets/android/AndroidDirect.kt
malliaridis: The android application plugin cannot be configured together with the multiplatform plugins. This is the migration requirement for AGP 9.

Just removing the "apply" here would simply not apply the plugin. This would probably break the rest of the configuration in this file. Upgrading AGP. to 9 in 7.0 will require users to also migrate their projects to not use android application plugin in the same module as the multiplatform shared code.

Providing a migration guide is required here.

Whether we need to configure the android executable function though is another topic. I would like to cleanup the korge-plugin, but in a separate move, and remove any configuration that we should not be countable for (e.g. project details that are configured in gradle per project).

This requires some better planning and investigation of what is configured and how in this plugin.

korge-gradle-plugins/korge-gradle-plugin/src/main/kotlin/korlibs/korge/gradle/targets/android/Android.kt
malliaridis: As mentioned in other comment, for higher compatibility, Java 21 may be the better option right now. But I am open to discuss and test an upgrade to 25 as well. :)

korge-gradle-plugins/korge-gradle-plugin/build.gradle.kts
malliaridis: If we had that before, then yes. But we may also move to a more "reusable" group name by using `org.korge.gradle` everywhere. But I will leave that open for further consideration.

of course we should align it either here or in the other places.

korge-gradle-plugins/korge-gradle-plugin-common/build.gradle.kts
malliaridis: I believe `org.korge.gradle` is a more "reusable" namespace, that could incolude anything related to gradle. But `gradleplugins` is fine too.

RezMike: I suggest using `plugin`/`plugins` word in the group name, because `org.korge.gradle` can be used for a separate korge library with the gradle support, while `org.korge.gradle.plugin` or `org.korge.gradleplugins` explicitly sets the correct scope

korge-gradle-plugins/korge-gradle-plugin/src/main/kotlin/korlibs/korge/gradle/targets/android/Android.kt
jobe-m: I guess I mixed it up with the need to use/update to Java 25 in Intellij IDEA/Android Studio because of Gradle 9.4.1. But here the Java version means something else, right? So, it's fine to go with Java 21 :)

korge-gradle-plugins/korge-gradle-plugin/src/main/kotlin/korlibs/korge/gradle/targets/android/AndroidDirect.kt
jobe-m: > Upgrading AGP. to 9 in 7.0 will require users to also migrate their projects to not use android application plugin in the same module as the multiplatform shared code.

Korge is "hiding" the Gradle stuff from the (game) developer. So I guess there are not many of us which will have any additional Gradle code outside of Korge in their game project configuration. I even remember that adding another dependency to a library from Maven Central (as I did for Fleks) required to "go" through Korge and use a provided Gradle plugin function `dependencyMulti("io.github.quillraven.fleks:Fleks:2.14", registerPlugin = false)`. So I guess there will not be the need for a migration guide.

If "com.android.application" cannot be used - as it seems - then let's remove it and add what is necessary to get Korge working again :)

gradle/libs.versions.toml
jobe-m: Agree

gradle/libs.versions.toml
jobe-m: `korge-sandbox` needs to be adjusted also. It uses Korge basically the same (?) way as a standalone Korge game project would use Korge. I guess it can be used to quickly test Korge features without the need of releasing Korge locally and using it in another project.

malliaridis: You may try the sandbox again and let me know if it works for JVM now. I have enabled and configured a few things to make it work again (tested on MacOS). There are basically 3 executable tasks:

- `jvmRun`
- `runJvm`
- `runJvmAwtSandbox`

korge-gradle-plugins/korge-gradle-plugin/src/main/kotlin/org/korge/kproject/KProjectPlugin.kt
malliaridis: The replacement for it is `KotlinMultiplatformAndroidLibraryExtension` now. I have updated the file accordingly, further deprecations need to be identified and updated. Thanks for finding this one. I will also take a look at other repos once this is no longer in draft state.

korge-gradle-plugins/korge-gradle-plugin/src/main/kotlin/korlibs/korge/gradle/targets/android/AndroidDirect.kt
malliaridis: Yeah, this is what I found problematic with the current plugin setup. We are basically "redefining" what gradle already does. We have like an alias for everything just to make it "the korge way", without any special reason. Sure there was or is a reason, but it can definitely be solved differently, more gradle-like, allowing users who already are familiar with gradle use their existing knowledge. So `korge` would be just another plugin that turns your project into a game, similar to what the compose plugin does (turning your project to a frontend multiplatform application).

korge-gradle-plugins/korge-gradle-plugin/src/main/kotlin/korlibs/korge/gradle/targets/android/Android.kt
malliaridis: The selection of the java version mainly depends on the libraries used in the game projects. When going with the lets say Java 25, any other dependency added to the project needs to support that version as well, otherwise it won't compile. As of today, I believe the highest compatibility is reached with 21, not 25 (yet).

I guess we can upgrade to 25 in the next major release though, or allow the user to configure it (best option) and only fallback to 21 if the user is not interested in setting it. But I would address that outside the scope of this PR.

korge-gradle-plugins/korge-gradle-plugin-common/build.gradle.kts
malliaridis: I added the version info here and removed it from the publishing configurations, as it would duplicate it for no reason.

korge-gradle-plugins/korge-gradle-plugin-settings/build.gradle.kts
malliaridis: I used gradleplugins everywhere for now. :)

korge-gradle-plugins/korge-gradle-plugin-common/build.gradle.kts
malliaridis: Makes sense. I updated the plugins group to `org.korge.gradleplugins` as suggested.

korge-gradle-plugins/korge-gradle-plugin/src/main/kotlin/korlibs/korge/gradle/targets/android/AndroidDirect.kt
jobe-m: This is a very good point. I would also like that Korge does not add its "own" Gradle stuff. This is hard to understand and to maintain. I am fine if we find a good solution for 7.0.0 (or 8?) which just uses Gradle in a minimalistic way. This also makes maintenance much easier.
When there is already a good approach (compose plugin) than why not utilizing it also for Korge.

gradle/libs.versions.toml
jobe-m: I could also run sandbox "runJvm" successfully on win11.

jobe-m:
I will publish this state locally and try to use it in my game. Unfortunately I will not have time next week so I might be a bit late with this task.

@malliaridis Feel free to merge this PR when it would enable you to to further with other PRs. I have reviewed it and it looks OK to me!

jobe-m:
augment review

jobe-m (approved):

jobe-m:
> I will publish this state locally and try to use it in my game. Unfortunately I will not have time next week so I might be a bit late with this task.
> 
> @malliaridis Feel free to merge this PR when it would enable you to to further with other PRs. I have reviewed it and it looks OK to me!

I did a quick check with my game (using kproject plugin) and still get similar errors as before on Gradle sync:

When Android target enabled: (with `targetAndroid()` inside `korge { ... }` in build.gradle.kts of the project)

```
* Where:
Build file 'C:\Users\jobe\devel\jobesLegacyGame\build.gradle.kts' line: 42

* What went wrong:
Failed to apply plugin 'com.android.internal.application'.
> The 'com.android.library' (or 'com.android.application') plugin is not compatible with the 'org.jetbrains.kotlin.multiplatform' plugin since AGP 9.0.
  Solution:
    - [Recommended] Replace the 'com.android.library' plugin with the 'com.android.kotlin.multiplatform.library' plugin (see https://developer.android.com/kotlin/multiplatform/plugin).
    - Or set the Gradle property 'android.builtInKotlin=false' and 'android.newDsl=false' to temporarily bypass this issue.
```

After disabling Android target:

```
An exception occurred applying plugin request [id: 'org.korge.kproject']
> Failed to apply plugin 'org.korge.kproject'.
   > Extension of type 'KotlinMultiplatformAndroidLibraryExtension' does not exist. Currently registered extension types: [ExtraPropertiesExtension, LibrariesForLibs, VersionCatalogsExtension, KotlinMultiplatformExtension, KotlinTestsRegistry, BasePluginExtension, SourceSetContainer, ReportingExtension, JavaToolchainService, JavaPluginExtension, KotlinArtifactsExtensionImpl, KotlinMultiplatformAndroidComponentsExtension]
```

Please don't mind my test for merging this PR. I would like to disable (or exclude) the kproject Gradle plugin completely for now from being published. I guess we need to remove it from the korge-gradle-plugin since when I remove all kproject related config from by game project than I still get those errors. But I am fine to add my Korge addon "Korge-fleks" as normal Gradle module to my project.

malliaridis:
I am currently updating the e2e tests where I face the same issue. I will fix it and let you know what needs to be updated and how. If removing kproject is fine, I will probably go and remove it from the plugin as well.

jobe-m:
> I am currently updating the e2e tests where I face the same issue. I will fix it and let you know what needs to be updated and how. If removing kproject is fine, I will probably go and remove it from the plugin as well.

I also removed it locally from registering in the plugin code. Then it was not published any more. With that state published locally I could successfully compile my game. But I also needed to disable Android. That was the last thing I could test. I will continue after my vacation :)
Feel free to remove it from the gradle plugin for now.

malliaridis:
Ok, so the current state of the e2e projects as well as the gradle plugins require some migration to a new project structure. I will address this in a separate PR by following the current recommendations from Jetbrains (see [this video](https://youtu.be/Atvl0l7fm1Y?si=WIAP38-YxnOKumaT)).

The main idea of the additional migration is to separate the entry points of android, ios, desktop and web to separate modules. This turns the "core" project into shared code, and is considered a library therefore. By applying only the library plugins of kotlin, korge is compatible with the AGP 9 requirements. However, this requires separate module for android, and for consistency, desktop and web as well. 7.0 release will require a migration guide for users, but it is definitely managable.

I will add to 7.0 roadmap a few tasks for that migration and for the simplification of the korge plugins.

jobe-m:
> Ok, so the current state of the e2e projects as well as the gradle plugins require some migration to a new project structure. I will address this in a separate PR by following the current recommendations from Jetbrains (see [this video](https://youtu.be/Atvl0l7fm1Y?si=WIAP38-YxnOKumaT)).
> 
> The main idea of the additional migration is to separate the entry points of android, ios, desktop and web to separate modules. This turns the "core" project into shared code, and is considered a library therefore. By applying only the library plugins of kotlin, korge is compatible with the AGP 9 requirements. However, this requires separate module for android, and for consistency, desktop and web as well. 7.0 release will require a migration guide for users, but it is definitely managable.
> 
> I will add to 7.0 roadmap a few tasks for that migration and for the simplification of the korge plugins.

I created an issue for this in "Korge 7 Community Release" project: https://github.com/korlibs/korge/issues/2369

jobe-m:
e2e test will be fixed with: https://github.com/korlibs/korge/issues/2366

PR #2374, opened 2026-07-04
chore: Cleanup korge gradle plugins

malliaridis:
## Description

This PR removes dead code and empty classes. Additionally, it replaces some deprecations in the plugin files, like `buildDir`. 

Since the `korge-gradle-plugin-common` is not containing any valuable content, it was fully removed as well and the only plugin left `korge-gradle-plugin` is flattened out.


jobe-m (approved):
Awesome! Thanks for cleaningh this all up!

PR #2375, opened 2026-07-05
chore: Remove korge plugin dependency configurations

malliaridis:
## Description

One of the current tasks of the korge gradle plugin is to apply a basic set of configurations, including dependencies. This comes however with certain limitations, additional complexity and maintenance cost. This behavior becomes redundant, as it is mostly adding a new way of configuring things like plugins and dependencies, which is limited and not the standard way of doing things.

## Solution

This PR removes most of the configurations that apply plugins and dependencies via the korge gradle plugin so that the users are in full control and have only the standard way to configure plugins and dependencies, being in full control of the versions that are included and the dependencies that are applied to individual targets.

## Breaking Changes

The following dependencies are no longer applicable or applied by the korge-gradle-plugin:

| Dependency | Configurations | API / Impl |
| -- | -- | -- |
| org.jetbrains.kotlinx:kotlinx-coroutines-core | commonMain | api |
| org.korge.engine:korge | commonMain | api |
| org.jetbrains.kotlin:kotlin-stdlib-common | commonMain | impl |
| org.jetbrains.kotlin:kotlin-test | commonTest | impl |
| org.jetbrains.kotlinx:kotlinx-serialization-json | commonMain | api |
| net.java.dev.jna:jna | jvmMain | impl |
| net.java.dev.jna:jna-platform | jvmMain | impl |
| org.jetbrains.kotlin:kotlin-stdlib-jdk8 | jvmMain | impl |
| org.jetbrains.kotlin:kotlin-test | androidTest | impl |
| androidx.test:core | androidTest | impl |
| androidx.test.ext:junit | androidTest | impl |
| androidx.test.espresso:espresso-core | androidTest | impl |

The following plugins are no longer applicable / applied by the korge-gradle-plugin:

| Plugin |
| -- |
| idea |
| [kotlin-multiplatform](https://kotlinlang.org/multiplatform/) |
| [kotlinx-serialization](https://github.com/kotlin/kotlinx.serialization) |

You should apply the plugins according to the plugin's documentation.

Additionally, the following functions and attributes are no longer available inside `korge {}`:

- dependencyMulti
- dependencyCInteropsExternal
- plugin
- addDependency
- dependencyProject
- configureAndroidDependency
- skipDevs
- serializationJson
- serialization

The following version references were also removed:

- BuildVersions.*
- jnaVersion
- korgeVersion
- kotlinVersion
- kotlinVersionIsDev
- androidBuildGradleVersion
- coroutinesVersion

Use a version catalog instead.

These functions were removed in favor to kotlin's / gradle's standard way of doing things.

Apply plugins and dependencies in the standard way. That is for KMP via `kotlin.sourceSets.[target].dependencies` and `plugins {}` in your `gradle.build.kts`.

Please note that it is also no longer supported to automatically apply gradle module `:deps` and skip it via `skipDevs = true`. Instead, you should use [version catalog bundles](https://docs.gradle.org/current/userguide/version_catalogs.html#sec:dependency-bundles).


malliaridis:
The next step is to address the drift of the plugin integration and module split in end-user applications, starting with the e2e projects inside korge. This PR already addresses a few issues, but does not solve the problems entirely.

jobe-m (approved):
LGTM!

jobe-m:
Tested successfully for JVM target on korge-hello-world, android compiles successfully but files from resources can still not be found (which is another error)

PR #2380, opened 2026-07-16
bugfix: Fix e2e test resources android

malliaridis:
## Description

While executing the e2e test example on Android, I noticed that the resources were not loaded properly and the screenshots were not captured.

Additionally, from #2375 we introduced a package missmatch for JVM.

## Solution

This PR addresses all these issues by using an app-specific external storage path, loads explicitly the commonMain resources as assets into the androidApp module, and updates the jvm main class package to match the correct one.

## Additional Information

The fix for loading the common resources as assets into the androidApp module should be only temporary and normally handled by the korge plugin, or at least moved to the shared module's logic. However, this require more time to investigate and figure out how to do it, so I chose the quick solution for now.

You may review the fixes by opening the screenshots on the test device. They can now be found under `/storage/emulated/0/Android/data/org.korge.e2e.test/files/`. Before the fix they were not created (errors were logged), and with both fixes at least the `Filters.png` is different and is rendering the images. I am not sure what the third ping screenshot is supposed to display.

### DIFF
--- a/e2e/e2e-test/androidApp/build.gradle.kts
+++ b/e2e/e2e-test/androidApp/build.gradle.kts
@@ -28,6 +28,13 @@ android {
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
     }
 
+    // TODO Consider moving this to the shared module or the korge plugin configurations
+    sourceSets {
+        getByName("main") {
+            assets.directories.add("../shared/src/commonMain/resources")
+        }
+    }
+
     // Ignore most of the configuration like packaging and signing, as we do not care in e2e tests
     packaging {
         resources {
--- a/e2e/e2e-test/androidApp/src/main/kotlin/org/korge/e2e/MainActivity.kt
+++ b/e2e/e2e-test/androidApp/src/main/kotlin/org/korge/e2e/MainActivity.kt
@@ -30,7 +30,7 @@ class MainActivity : KorgwActivity(config = GameWindowCreationConfig(msaa = 1, f
                 )
 
                 Log.d(TAG, "Determining screenshotsVfs...")
-                val path = Environment["OUTPUT_DIR"] ?: "."
+                val path = applicationContext.getExternalFilesDir(null)!!.absolutePath
                 Log.d(TAG, "Determining screenshotsVfs... $path")
                 val screenshotsVfs = localVfs(path).also { it.mkdirs() }.jail()
                 Log.d(TAG, "vfs: $screenshotsVfs")
--- a/e2e/e2e-test/desktopApp/build.gradle.kts
+++ b/e2e/e2e-test/desktopApp/build.gradle.kts
@@ -12,7 +12,7 @@ korge {
 
     targetJvm()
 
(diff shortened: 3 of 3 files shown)


e2e/e2e-test/androidApp/build.gradle.kts
```diff
@@ -28,6 +28,13 @@ android {
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
     }
 
+    // TODO Consider moving this to the shared module or the korge plugin configurations
```
jobe-m, line 31: I guess I have exactly that solved in the korge-hello-world example.
-> https://github.com/korlibs/korge-hello-world/pull/37/changes#diff-cf9ae3b1eb7273a914f22b1e6d22d8447f88df0e7245a5b856fb751e507b979bR10

I also realised that the kotlin multiplatform plugin installs all files from `src/commonMain/resources` into the root folder of an APK. While all "assets" files are in an assets subfolder in the APK, which lead to duplicate resource files in the APK.

Thus, in the example I moved the resources out of commonMain and placed it in the root of the shared module. Korge did something similar before and did also not use the "standard" resource path.

malliaridis: Is this duplication caused by kotlin multiplatform plugin, or by the korge plugin? I read somewhere that, at least for Compose resources, when placing resources inside `composeResources` it will be taken care of to put everything into the android `assets`, where they should be. This way it is possible to also ship large games (4GB+) via Play Store in chunks. Haven't released an APK of this size before, but I believe there is some digging needed here.

These lines of code should add the common resources into the android assets, at least that's what it is supposed to do. And I believe korlibs loaded the assets successful (without further changes) if I judge by the outcome. The only thing is that it should be handled ideally by the korge plugin or by a library, for any target.

I also am familiar with multiple resource loading libraries, I would probably try to avoid any of these if possible.

jobe-m: Yes, duplication is because of kotlin multiplatform plugin, not the Korge plugin.
Also compse multiplatform uses "another" location called `composeResources` because of that.

Good idea with looking into how compose handles resources, but due to limited time here I will not do it. Also it is questionable that a "hobby" project like my game would ever exceed 4GB assets ;-)

Currently the easiest is to specify the resource folder for assets as a separate folder under "shared" and configure it to be copied/used by the targets.

jobe-m (approved):
LGTM!

e2e/e2e-test/androidApp/build.gradle.kts
jobe-m: In the end it does not matter where the config for specifying the assets folder for Android is done. So let's use this implementation.

malliaridis: I think for this PR it is fine as it is, but it is definitely an issue worth for researching and implementing a proper way with the korge plugin. Perhaps a lightweight library is also a good option.

malliaridis: Added this issue for tracking: https://github.com/korlibs/korge/issues/2382

PR #2381, opened 2026-07-16
chore: Remove deprecations and dead code

malliaridis:
## Description

Since we are using now the default kotlin multiplatform hierarchy for sourcesets, a lot of code still can be removed or simplified.

## Solution

This PR contains three changes:
- Remove the configurations that use the old structure like `src@[target]` and `test@[target]` and rely solely on the new structure
- Remove `RootKorlibsPlugin` together with `ConfigureKover` as it is not used anywhere
- Remove any remaining references to `macosX64`, a target that is no longer supported


korge-core/src/commonTest/kotlin/korlibs/render/GameWindowTest.kt
malliaridis, line 1: This one just slipped into the commits, was not intentional but we can keep it.

jobe-m (approved):
LGTM

PR #2387, opened 2026-08-12
bugfix: Fix swiftui workflow by selecting arm64 architecture explicitly

malliaridis:
## Description

For a while the SwiftUI integration test failed, likely because of the wrong architecture being selected / included.

## Solution

Configure the workflow step to link only for arm64 in xcodebuild.

Additionally, change task `build` to `check` to avoid executing long-lasting tasks like `linkRelease*`. This reduces the build time significantly from ~35min to ~11min.

## Additional Notes

This PR also adds yarn.lock back to repository, as these files have to be part of version-control (otherwise tasks like kotlinNpmInstall may fail (see checks).


jobe-m (approved):

jobe-m:
Great, thank you for fixing this :)

